PR #3, opened 2015-09-21
Update travis config

ADmad:
Travis' webhook needs to be enabled for the repo.


markstory:
I double checked - webhook is on now.

PR #7, opened 2015-09-23
update .gitattributes for file types

dereuromark:
Similar to CakePHP core.
Makes the newline conversion correct for all OS.


antograssiot:
I was about to make it :wink:

PR #8, opened 2015-09-25
Fix a pile of PHPCS errors.

markstory:
The `$thing::class` errors need to be fixed in the codesniffer I think.


src/Chronos.php
```diff
@@ -98,7 +98,7 @@ public function toMutable()
     /**
      * Get a copy of the instance
      *
-     * @return static
+     * @return $this
```
dereuromark: Phpstorm understands now `self` as well b the way.

PR #9, opened 2015-09-26
Update more docs.

markstory:
Update more docs. Almost have a passing PHPCS build.


antograssiot:
I added a small correction as a commit. looks good.

PR #11, opened 2015-09-26
Passing CS build.

bcrowe:


tests/DateTime/SettersTest.php
```diff
@@ -103,10 +103,10 @@ public function testTimeSetterWithChaining()
     {
         $d = MutableDateTime::now();
         $d->setTime(2, 2, 2)->setTime(1, 1, 1);
-        $this->assertInstanceOf(MutableDateTime::class, $d);
+        $this->assertInstanceOf(MutableDateTime::CLASS, $d);
```
lorenzo: This does not look right

markstory: The [PHP docs](http://php.net/manual/en/migration55.new-features.php#migration55.new-features.class-name) have this as lowercase. I was thinking of updating the code-sniffer rules (ours or squizlabs) so that `::class` passed.

PR #15, opened 2015-09-29
Start splitting the mega trait into more managable pieces

markstory:
This splits out comparison methods and string formatting methods.


lorenzo:
Good job!

PR #18, opened 2015-09-29
Made addMonth() and subMonth() not overflow

### LINKED ISSUE #10: Make subMonth work correctly
See https://github.com/briannesbitt/Carbon/issues/428

I think it is worth fixing the faulty (unintuitive) behavior of PHP, and making subMonth() work as expected.
To make sure the former behavior can also be used if desired, we could provide a method `addMonthsExact()` or `addMonthsOverflow()`.

What do you think?
Since the lib is new we do have the change to do that without breaking user land code.

lorenzo:
Fixes #10

### DIFF
--- a/src/DateTimeTrait.php
+++ b/src/DateTimeTrait.php
@@ -914,7 +914,16 @@ trait DateTimeTrait
      */
     public function addMonths($value)
     {
-        return $this->modify((int)$value . ' month');
+        $date = $this->copy()->modify((int)$value . ' month');
+
+        if ($date->day != $this->day) {
+            return $date
+                ->day(1)
+                ->subMonth()
+                ->endOfMonth();
+        }
+
+        return $date;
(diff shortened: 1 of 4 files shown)


dereuromark:
:+1:

src/DateTimeTrait.php
```diff
@@ -909,17 +909,48 @@ public function subYears($value)
...
      * @return static
      */
     public function addMonths($value)
     {
-        return $this->modify((int)$value . ' month');
+        $day = $this->day;
+        $date = $this->modify((int)$value . ' month');
+
+        if ($date->day !== $day) {
+            return $date->modify('last day of previous month');
```
markstory: whooah, I didn't know this existed.

markstory:
Looks good to me. I fixed some typos in the docs blocks.

Issue #12, opened 2015-09-27
RFC - Split up DateTimeTrait into smaller traits


markstory:
The current DateTimeTrait is over 2000 lines long. I was wondering if people would be up to splitting up the mega trait into a few smaller traits. I see a few types of methods:

* Formatting methods. (toRssString) etc.
* Modifiers (endOfWeek, startOfDay, addSecond, addMinute, lastOfMonth) etc.
* Comparisons (isToday, isThisYear, isThursday) etc.
* Operators (eq, neq, lt, lte, gt, gte)
* Timezones (tz, timezone)

I think splitting up the mega trait will help with composability when we make a calendar date object.

dereuromark:
Great idea!

lorenzo:
Makes sense to me

markstory:
Finished with #21

PR #19, opened 2015-10-02
Extract more traits

markstory:
I've split out 3 more traits, and cleaned up the interface a bit.

I felt that having static methods in the interface was of little/no value, but I'm happy to put them back if others disagree.

Refs #12


antograssiot:
I've removed some extra use statements. I can re-add them if you think it increase dependencies readability.

dereuromark:
Only one remark: All traits are singular "Timezone", "Comparison", etc.
I think the "Modifiers" trait should be named "Modifier" Trait as well.

markstory:
@dereuromark Good point. I will fix that up.

jadb:
Nice catch @dereuromark!

PR #24, opened 2015-10-11
Allow isBirthday() to work with current day by default.

dereuromark:
Curently you would need to pass an object as argument, whereas intuitively you shouldn't have to if you are referring to the present (which is the default IMO).

So it simplifies to:
```php
$birthdayString = $this->request->data['birthday']['year'] . '-' . $this->request->data['birthday']['month'] . '-' . $this->request->data['birthday']['day'];

$birthday = new Chronos($birthdayString);
$isBirthday = $birthday->isBirthday(); // Today
```

### DIFF
--- a/src/ComparisonTrait.php
+++ b/src/ComparisonTrait.php
@@ -358,11 +358,14 @@ trait ComparisonTrait
     /**
      * Check if its the birthday. Compares the date/month values of the two dates.
      *
-     * @param ChronosInterface $dt The instance to compare with.
+     * @param ChronosInterface|null $dt The instance to compare with or null to use current day.
      * @return static
      */
-    public function isBirthday(ChronosInterface $dt)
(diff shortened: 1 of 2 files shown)


src/ComparisonTrait.php
```diff
@@ -358,11 +358,14 @@ public function isThisYear()
...
      *
-     * @param ChronosInterface $dt The instance to compare with.
+     * @param ChronosInterface|null $dt The instance to compare with or null to use current day.
      * @return static
      */
-    public function isBirthday(ChronosInterface $dt)
+    public function isBirthday(ChronosInterface $dt = null)
     {
+        if ($dt === null) {
+            $dt = static::now($this->tz);
```
markstory: We should have a test. :)

dereuromark: Test added.

PR #27, opened 2015-10-11
Correct doc blocks and argument defaults.

dereuromark:


src/ChronosInterval.php
```diff
@@ -134,13 +134,13 @@ public function __construct($years = 1, $months = null, $weeks = null, $days = n
...
      * (new ChronosInterval(1))->fn().
      *
-     * @param int $years The year to use.
-     * @param int $months The month to use.
-     * @param int $weeks The week to use.
-     * @param int $days The day to use.
-     * @param int $hours The hours to use.
-     * @param int $minutes The minutes to use.
-     * @param int $seconds The seconds to use.
+     * @param int|null $years The year to use.
```
bcrowe: `$years` defaults to 1

dereuromark: Yes, I know. But you can pass (and you do) null as well, see the rest of the call code. So it has to be consistent here, as well.

Issue #33, opened 2015-10-12
Allow for more expressive comparisons


dereuromark:
It's easier to quickly understand a line of code by reading `->greaterThan()` than `->gt()`.

I think we should use the non abbr version, and remove the shortcuts.
Less obfuscation.

Refs https://github.com/briannesbitt/Carbon/pull/446

So renaming:

    eq() => equalTo()
    ne() => notEqualTo()
    gt() => greaterThan()    
    lt() => lessThan()
    gte() => greaterThanOrEqualTo()    
    lte() => lessThanOrEqualTo()

It also fits to verbose naming inside cakephp core and other projects like phpunit.

optionally drop aliases for minimum/maximum/timezone:

    min() => already exists as minimum()
    max() => already exists as maximum()
    tz() => already exists as setTimezone() and timezone()!

having too many aliases for no reason is usually sth we are used to from PHP core but shouldn't go down the same rabbit hole IMO.

markstory:
Closing as there is a pull request open now.

PR #36, opened 2015-10-13
Allow for more expressive comparisons and remove unneeded aliases.

dereuromark:
Resolves https://github.com/cakephp/chronos/issues/33

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -137,14 +137,6 @@ interface ChronosInterface extends DateTimeInterface
     public function timezone($value);
 
     /**
-     * Alias for setTimezone()
-     *
-     * @param DateTimeZone|string $value The DateTimeZone object or timezone name to use.
-     * @return static
-     */
-    public function tz($value);
-
-    /**
      * Set the instance's timezone from a string or object
      *
      * @param DateTimeZone|string $value The DateTimeZone object or timezone name to use.
@@ -270,7 +262,7 @@ interface ChronosInterface extends DateTimeInterface
      * @param ChronosInterface $dt The instance to compare with.
      * @return bool
      */
-    public function eq(ChronosInterface $dt);
+    public function equalTo(ChronosInterface $dt);
 
     /**
      * Determines if the instance is not equal to another
@@ -278,7 +270,7 @@ interface ChronosInterface extends DateTimeInterface
      * @param ChronosInterface $dt The instance to compare with.
      * @return bool
      */
-    public function ne(ChronosInterface $dt);
+    public function notEqualTo(ChronosInterface $dt);
 
     /**
      * Determines if the instance is greater (after) than another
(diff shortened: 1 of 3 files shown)


src/ChronosInterface.php
```diff
@@ -336,7 +328,7 @@ public function min(ChronosInterface $dt = null);
      * @param ChronosInterface|null $dt The instance to compare with.
      * @return static
      */
-    public function max(ChronosInterface $dt = null);
+    public function maximum(ChronosInterface $dt = null);
```
markstory: min/max have equivalent functions in PHP stdlib.

dereuromark: So you think we should keep those?

lorenzo:
I'm not sold on this. One of the goals of Chronos was being API compatible with Carbon so it could be replaced. Also, some of the names like `lte` and `gt` are already function names in the ORM. Personally I think people are used to those.

dereuromark:
We can leave the old methods in as `@deprecated` for the transition to be smoother (and remove those in v2.0).
But I would rather have clear and descriptive methods than cryptic shortcuts that are not easy to read.

ADmad:
I too am not sold on this. These short names are pretty common and most would know what they mean.

markstory:
The short names also happen to be the same as those used in python for operator overloading (minus underscores).

dereuromark:
I wonder how common that knowledge is in the PHP world :)
But all right, staying then it seems.

burzum:
Can't we just have both? :)

markstory:
@burzum I'm fine with more aliases as well.

rchavik:
I'd rather not. Let's pick one and stick with it.

ADmad:
I am not a fan of multiple names for the same thing either.

dereuromark:
Hm, with https://github.com/briannesbitt/Carbon/pull/512/files being almost merged now we need to also consider adding them then (instead of replacing, we will have additional aliases).

IMO we should at least deprecate the ones we do not want to continue beyond 1.x.
This way we can easier clean up then in a future 2.x version.

dereuromark:
This is merged now, btw: https://github.com/briannesbitt/Carbon/pull/512
Shall we then include them and mark them as `@deprecated` for the next major version?

ADmad:
I am :-1: for marking shorter methods names as deprecated. Would rather live with aliasing.

dereuromark:
I meant the longer ones (which have been added just now). We don't need to have all aliases in the future.

markstory:
Having multiple aliases doesn't seem that bad. It lets people use the style they want which is nice. We can revisit whether or not we drop one set of aliases in the future I think.

dereuromark:
OK, then we should add the aliases that Carbon introduced recently into this library for BC reasons, as well.

ionas:
If chronos is meant to be a carbon replacement in future / out of cakephp BC scope. Yes.

Else my preferences are:
1. Long versions (deprecate shortcuts)
2. Short versions (deprecate long versions)
3. Having both

dereuromark:
We probably now have to go with both, someone just needs to do the work and update it to a new PR.

ionas:
I'd rather stick to either being fully carbon compatible or reduce the set of methods to one.
https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Carbon.php#L792

dereuromark:
Deprecating sth is fully compatible still.

Issue #13, opened 2015-09-27
RFC - Add a Calendar date object


markstory:
One of the issues we've heard a bunch of times with the ORM is that DATE columns are mapped to datetime instances. While we can't completely avoid that, we could make a 'calendar' date object. This object would differ from the datetime instance in a few ways:

* The time methods would be removed or raise errors.
* The time mutators would raise errors or not exist. Including the methods and raising exceptions lets us conform to the ChronosInterface which is nice.
* On construction the time component would be set to `00:00:00`.

HavokInspiration:
Good idea.
I often had to manipulate "calendar" date and having something dedicated to that would be super nice.

As for the various points, I'd rather have errors raised rather than methods removed. I think that would avoid confusion to have a clear error message stating that the methods can't be used and that if the "time component" is needed, the wrong object is being used.

lorenzo:
:+1:

markstory:
@HavokInspiration Ok. Given the scope of the chronosinterface, we'll need to raise exceptions anyways.

HavokInspiration:
@markstory All good then ! (I just checked the ChronosInterface... It is quite lengthy).

robertpustulka:
BTW It's rather not very OOP to have interfaces over 1000 lines long...

lorenzo:
I'm not sure it is a good idea to throw exceptions in those methods. If we do then we need to make every other class aware of the differences between a normal datetime and a calendar date.

markstory:
@robertpustulka DateTimes are all knowing all powerful objects :wine_glass:

robertpustulka:
@markstory I was thinking that maybe we could split ChronosInterface into TimeInterface and CalendarInterface, but what about compatibility with DateTimeInterface then? 
Need to remind myself that PHP5 is not very OOP either :)

markstory:
Breaking interfaces up might be an option. Once I'm done splitting the megatrait up, I'll get started on this.

dereuromark:
How about having Time and Date then as classes?

```php
use Cake\Chronos\Time;
use Cake\Chronos\Date;

$time = new Time();
$date = new Date();
```
so Chronos class itself => rename to Time, adding Date, both extending \DateTime then.
That would also get rid of the double name in `use Cake\Chronos\Chronos` and make it more clear what those two splits refer to.

markstory:
[...]

PR #40, opened 2015-10-20
RFC - Rough implementation of Calendar Date

markstory:
This is still somewhat rough, but I wanted to get some feedback on the direction and some of the assumptions I've made.

### Assumptions

* Calendar Dates cannot have times added/mutated.
* Calendar Dates are immutable. There will be no mutable version of this object type.
* Calendar Dates can reasonably be normalized to UTC. Given that each calendar date occurs in all timezones, picking one to normalize on is fairly arbitrary.

### Things left to do

* [x] Hunt down other mutator methods that change times and nerf them.
* [x] Implement toString

Related to #13

### DIFF
--- a/src/Date.php
+++ b/src/Date.php
@@ -34,6 +34,67 @@ class Date extends DateTimeImmutable implements ChronosInterface
     use TestingAidTrait;
 
     /**
+     * Create a new Immutable Date instance.
+     *
+     * Please see the testing aids section (specifically static::setTestNow())
+     * for more on the possibility of this constructor returning a test instance.
+     *
+     * Date instances lack time components, however due to limitations in PHP's
+     * internal Datetime object the time will always be set to 00:00:00, and the
+     * timezone will always be UTC. Normalizing the timezone allows for 
+     * subtraction/addition to have deterministic results.
+     *
+     * @param string|null $time Fixed or relative time
+     * @param DateTimeZone|string|null $tz The timezone for the instance
+     */
+    public function __construct($time = null, $tz = null)
+    {
+        $tz = new DateTimeZone('UTC');
+        if (static::$testNow === null) {
+            $time = $this->stripTime($time);
+            return parent::__construct($time, $tz);
+        }
+
+        $relative = static::hasRelativeKeywords($time);
+        if (!empty($time) && $time !== 'now' && !$relative) {
+            $time = $this->stripTime($time);
+            return parent::__construct($time, $tz);
+        }
+
+        $testInstance = static::getTestNow();
+        if ($relative) {
+            $testInstance = $testInstance->modify($time);
+        }
+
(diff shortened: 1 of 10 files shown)


lorenzo:
The general direction looks good. You also need to add to the TODO list something about the default toString result.

src/Date.php
```diff
@@ -0,0 +1,189 @@
...
+        return preg_replace('/\d{1,2}:\d{1,2}:\d{1,2}/', '00:00:00', $time);
+    }
+
+    /**
+     * No-op method.
+     *
+     * Timezones have no effect on calendar dates.
+     *
+     * @param DateTimeZone|string $value The DateTimeZone object or timezone name to use.
+     * @return static
```
dereuromark: Is static still true? Should that be self or $this?

src/Date.php
```diff
@@ -0,0 +1,189 @@
...
+     */
+    public function tz($value)
```
dereuromark: Can we deprecated tz() and not reimplement it as there is already timezone() and setTimezone() as two aliases?

markstory: I'm pretty sure tz() is part of the ChronosInterface.

markstory:
@lorenzo I added `__toString()` to the TODOs. I hope to take care of those things tonight.

markstory:
@lorenzo Default __toString() implemented and tests added for the other string methods.

tests/Date/ConstructTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Data provider for constructor testing.
+     *
+     * @return array
+     */
+    public function inputTimeProvider()
+    {
+        return [
+            [strtotime('2015-10-19 22:24:32')],
```
lorenzo: I think `DateTime` can only receive integers in the format `@12345`

markstory: I wonder how this is working then. I'll double check it.

markstory: Fixed now.

markstory:
In thinking some more and looking at other mutator methods, having methods that effect time throw exceptions really limits the utility of the ModifierTrait and this new class. Instead of throwing exceptions it seems more useful to always reset time to 00:00:00. This approach lets the fluent interface be used with any ChronosInterface and Date/DateTime can be used interchangeably without errors or compromising object integrity.

lorenzo:
@markstory I completely agree. I was looking at some of the uses of time objects inside the core and I could see places where those exceptions were going to be problematic

lorenzo:
This is looking great! Are there any other mutators that need to be hijacked?

markstory:
@lorenzo I don't think there are any more mutators. By attacking the ones provided by DateTimeImmutable, I think we are covered.

lorenzo:
Great job, @markstory

PR #41, opened 2015-10-26
More documentation

markstory:
Add some more documentation including a migration guide from carbon, and links to the API docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ echo $today;
 // Outputs '2015-10-21'
 
 echo $today->modify('+3 hours');
-// Outputs '2015-12-21'
+// Outputs '2015-10-21'
 ```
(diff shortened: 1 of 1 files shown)


josegonzalez:
@markstory you can add `[ci skip]` to skip ci on stuff like this. Not sure if we necessarily need it on docs changes...

README.md
```diff
@@ -5,4 +5,104 @@
...
+
+```php
+use Cake\Chronos\Date;
+
+$today = new Date();
+echo $today;
+// Outputs '2015-10-21'
+
+echo $today->modify('+3 hours');
+// Outputs '2015-12-21'
```
bcrowe: Should the month be changing to 12 here?

PR #43, opened 2015-10-28
Add methods from Cake\I18n\Time

markstory:
I started to integrate chronos into CakePHP, and found a few methods that were defined on `Cake\I18n\Time` that don't really need to be there anymore. By moving these methods we add more features to this library, but allow the I18n subclasses to be more focused on i18n extensions.


antograssiot:
good idea :+1:

PR #46, opened 2015-11-02
Port behavior from CakePHP in.

markstory:
The Cake\I18n\Time has this bizarre behavior, and since I'd rather avoid causing possibly breaking changes I figured it made sense to add this here. Without this change we will end up failing ~10 tests in the TimeHelper suite.

### DIFF
--- a/src/ComparisonTrait.php
+++ b/src/ComparisonTrait.php
@@ -378,6 +378,10 @@ trait ComparisonTrait
      */
     public function wasWithinLast($timeInterval)
     {
+        $tmp = trim($timeInterval);
+        if (is_numeric($tmp)) {
+            $timeInterval = $tmp . ' days';
+        }
         $now = new static();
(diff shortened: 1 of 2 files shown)


ADmad:
This is sad.

lorenzo:
I had removed that on purpose since it was not documented and made absolutely no sense

markstory:
@lorenzo Are we ok with there being a breaking change in 3.2 then? I'm also open to adding this goofy behavior into the CakePHP subclass.

josegonzalez:
I would rather put the goofy behavior in the cakephp subclass than here.

dereuromark:
CakePHP subclass + some maybe a note that it must be removed for 4.0 then.

lorenzo:
@markstory I'm ok with either breaking or subclassing. I don't think the goofy behavior was actually documented

markstory:
Ok, moving to CakePHP subclass then.

Issue #49, opened 2015-11-04
Port: setTimeFromTimeString()


dereuromark:
To allow easier porting for those coming from Carbon, it is probably wise to include
https://github.com/briannesbitt/Carbon/pull/452

    setTimeFromTimeString($time)

dereuromark:
PR is now open.

PR #50, opened 2015-11-06
Implement setTimeFromTimeString()

dereuromark:
Resolves https://github.com/cakephp/chronos/issues/49

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -119,7 +119,7 @@ interface ChronosInterface extends DateTimeInterface
      */
     public function setDateTime($year, $month, $day, $hour, $minute, $second = 0);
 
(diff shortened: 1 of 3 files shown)


lorenzo:
Would you mind adding the method to the interface?

dereuromark:
No problem.

ADmad:
Can we have the method name be shorter "setTimeFromString"?

dereuromark:
Sounds good to me, but bear in mind that this would make Carbon transition non BC then for anyone "upgrading" to ours later... So for API consistency we should not do that at this point as we don't do for other useful changes.

lorenzo:
:+1:

dereuromark:
Squashed.

PR #53, opened 2015-11-12
Reorganize code layout.

markstory:
Instead of one big directory for all the things, add a subdirectory for all the traits. I feel this keeps the library a bit more structured and understandable.


markstory:
Thanks @ADmad

PR #55, opened 2015-11-13
Add Mutable date object.

markstory:
A mutable date object is necessary for backwards compatibility in CakePHP. Without this class we cannot provide backwards compatible typehints and behaviors for Date column types.


markstory:
I can look into how to bring the coverage back up tomorrow.

markstory:
Coverage issues should be fixed now.

PR #56, opened 2015-11-14
Add class aliases for Carbon.

markstory:
This should help make switching libraries even easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ php composer.phar require cakephp/chronos
 
 By default Chronos includes a compatiblity script that creates aliases for the
 relevant Carbon classes.  This will let most applications upgrade with very
-little effort. However, if you'd like to permanently update your code,you will
+little effort. If you'd like to permanently update your code, you will
 need to update imports and typehints. Assuming `src` contains the files you
 want to migrate, we could use the following to update files:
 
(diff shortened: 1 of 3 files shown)


src/carbon_compat.php
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @copyright     Copyright (c) Brian Nesbitt <brian@nesbot.com>
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+// Create class aliases for Carbon so applications
+// can upgrade more easily.
+class_alias('Cake\Chronos\Chronos', 'Carbon\Carbon');
```
lorenzo: Let's alias it to MutableDateTime instead

README.md
```diff
@@ -36,9 +36,11 @@ First add `cakephp/chronos` to your `composer.json`:
 php composer.phar require cakephp/chronos
 ```
 
-To update your code you will need to update imports and typehints. Assuming
-`src` contains the files you want to migrate, we could use the following to
-update files:
+By default Chronos includes a compatiblity script that creates aliases for the
+relevant Carbon classes.  This will let most applications upgrade with very
+little effort. However, if you'd like to permanently update your code,you will
```
dereuromark: space after comma ", you"

src/carbon_compat.php
markstory: Good idea.

markstory:
Should be good now.

PR #57, opened 2015-11-19
Implement diffForHumans

markstory:
This shores up a big incompatibility between Carbon and Chronos. I've only implemented english, as I feel that is the minimal set of messages we need for compatibility. I also just didn't want to add any dependencies to chronos.

Refs cakephp/cakephp#7713

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -862,10 +862,10 @@ interface ChronosInterface extends DateTimeInterface
      * 5 months after
      *
      * @param \Cake\Chronos\ChronosInterface|null $other The datetime to compare with.
-     * @param bool $absolute removes time difference modifiers ago, after, etc
+     * @param bool $absolute Removes time difference modifiers ago, after, etc
      * @return string
      */
-    public function diffForhumans(ChronosInterface $other = null, $absolute = false);
+    public function diffForHumans(ChronosInterface $other = null, $absolute = false);
 
     /**
      * Get the difference in years
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -843,6 +843,31 @@ interface ChronosInterface extends DateTimeInterface
     public function subSeconds($value);
 
     /**
+     * Get the difference in a human readable format in the current locale.
+     *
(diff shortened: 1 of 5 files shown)


src/Traits/DifferenceTrait.php
```diff
@@ -234,4 +237,98 @@ public static function fromNow($datetime)
...
+        if (static::translator()->exists($tryKeyExists)) {
+            $time = static::translator()->plural($tryKeyExists, $count, array('count' => $count));
+        }
+        return static::translator()->singular($transId, array('time' => $time));
+    }
+
+    /**
+     * Get the translator instance or overwrite the current one.
+     *
+     * @return Cake\Chronos\Translator The translator instance.
```
dereuromark: |null

markstory: Thanks.

HavokInspiration:
This is looking good.
I fixed a bunch of CS warnings.

How do you suggest we provide other languages than english ?

lorenzo:
I was more inclined to see the implementation of the diff inside another object. Do you think that would make sense?

src/ChronosInterface.php
```diff
@@ -843,6 +843,31 @@ public function subSecond($value = 1);
...
+     *
+     * When comparing a value in the future to another value:
+     * 1 hour after
+     * 5 months after
+     *
+     * @param \Cake\Chronos\ChronosInterface|null $other The datetime to compare with.
+     * @param bool $absolute removes time difference modifiers ago, after, etc
+     * @return string
+     */
+    public function diffForhumans(ChronosInterface $other = null, $absolute = false);
```
HavokInspiration: Shouldn't it be ``diffForHumans`` ?

dereuromark: It should.
I would also uppercase the removes: `bool $absolute Removes`

markstory:
@lorenzo Ok. That might make it easier to integrate CakePHP's i18n subsystem.

markstory:
@lorenzo I've split out the formatter, let me know what you think.

lorenzo:
Good job @markstory!

Issue #54, opened 2015-11-12
addWeekday(s) clears time


dereuromark:
Check if
https://github.com/briannesbitt/Carbon/commit/d02c4136c3300e5044226e5be8fcfc5b4b6fc272
needs to be applied

Refs https://github.com/briannesbitt/Carbon/issues/419

lorenzo:
@dereuromark would you like to copy the fix from them?

dereuromark:
It seems that their fix only was for addWeekday. addWeekdays as well as the sub methods also need fixing i guess.
Additionally, the fixes cannot be applied without some modification for us: https://github.com/cakephp/chronos/compare/master-weekday
Tests failing.

markstory:
@dereuromark Those changes seem reasonable. For the immutable objects, you'll need to capture the return of `modify()` and chain off of that instead.

dereuromark:
You are absolutely right. Done :)

PR #58, opened 2015-11-22
Fix: addWeekday(s) clears time

dereuromark:
Resolves https://github.com/cakephp/chronos/issues/54

what about subWeekday and subWeekdays?
Those havent been fixed in carbon yet, but IMO should also be fixed.

### DIFF
--- a/src/Traits/ModifierTrait.php
+++ b/src/Traits/ModifierTrait.php
@@ -470,9 +470,20 @@ trait ModifierTrait
      */
     public function addWeekday($value = 1)
     {
+        return $this->addWeekdays($value);
+    }
+
+    /**
+     * Remove weekdays from the instance
+     *
+     * @param int $value The number of weekdays to remove.
+     * @return static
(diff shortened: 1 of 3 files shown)


dereuromark:
I now also fixed up the sub methods. But it seems there is some other bug occurring now: 
```
1) Cake\Chronos\Test\DateTime\SubTest::testSubWeekdaysPositive with data set "mutable" ('Cake\Chronos\MutableDateTime')
Exception: DateTimeZone::__construct(): Unknown or bad timezone (13)
```

markstory:
Code looks good. Do the tests also fail for you locally?

dereuromark:
Yes. Could this be a PHP bug with going back over the 1.1.?

dereuromark:
So the CS error is a regression in their sniffer: https://github.com/squizlabs/PHP_CodeSniffer/issues/786
Looks like this is OK to be merged then.

PR #59, opened 2015-11-28
Allow interval to be formatted as string.

dereuromark:
Implements https://github.com/briannesbitt/Carbon/pull/526 as __toString()

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -400,4 +400,39 @@ class ChronosInterval extends DateInterval
 
         return $this;
     }
+
+    /**
+     * Returns the ISO 8601 interval string.
+     *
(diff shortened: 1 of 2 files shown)


src/ChronosInterval.php
```diff
@@ -400,4 +400,39 @@ public function add(DateInterval $interval)
...
     }
+
+    /**
+     * Returns the ISO 8601 interval string.
+     *
+     * @return string Interval as string representation
+     */
+    public function __toString()
+    {
+        $date = array_filter(array(
```
antograssiot: short array syntax maybe here and bellow on line 417

dereuromark:
CS looks good now.

antograssiot:
:+1:

PR #60, opened 2015-11-28
Master coverage

dereuromark:
Write tests for so far untested functionality => improves coverage to almost 100%.


lorenzo:
Thanks!

PR #61, opened 2015-11-28
correct typos and remove unused statements

antograssiot:


ADmad:
Could you please take care of the CS errors too.

antograssiot:
@ADmad I removed the one that I added by mistake, for the remaining CS we need to wait for PHP_CodeSniffer 2.4.1 to be released

PR #63, opened 2015-12-05
Add tests against windows environment

antograssiot:
via appveyor like other part of the project.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,6 +28,8 @@ install:
   - cd C:\projects\chronos
   - php -r "readfile('https://getcomposer.org/installer');" | php
   - php composer.phar install --prefer-dist --no-interaction
+  - php -v
+  - php -i | grep "ICU version"
 test_script:
   - cd C:\projects\chronos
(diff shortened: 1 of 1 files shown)


markstory:
Looks good. We'll need to enable the integration once this merges.

markstory:
I think the existing cakephp integration is under @lorenzo 's account.

lorenzo:
I can take care of this and see if I can grant acces to the rest of the team

antograssiot:
@markstory @lorenzo for the record, right now the tests are failing on appveyor.
https://ci.appveyor.com/project/antograssiot/chronos/build/1.0.2

from the quick search that I've done it's due to this bug in PHP that has been fixed in 5.5.8
https://bugs.php.net/bug.php?id=65768

if I'm right, then appveyor is failling because it's run against 5.5.7 and it would probably require us to bump the minimal php version to 5.5.8 instead of 5.5.0 in composer.json and then to update again the php version used from ci.cakephp.org to make the tests.

PR #64, opened 2015-12-06
bump minimum php requirement to 5.5.8

antograssiot:
to avoid tickets due to https://bugs.php.net/bug.php?id=65768

If you agree with this change I can also update cakephp/cakephp and cakephp/app accordingly

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,13 @@
       "homepage": "http://cakephp.org"
     }
   ],
+  "support": {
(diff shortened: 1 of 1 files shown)


dakota:
Bumping cakephp/cakephp would require a major version bump no?

antograssiot:
@Dakota I don't see a huge difference in going to 5.5.0 or to 5.5.8 for 3.2 as planed.

lorenzo:
I think it makes sense

PR #65, opened 2015-12-06
Make diffFormatter() static.

markstory:
It operates on a static property, so having the method static makes sense.

Refs cakephp/cakephp#7789


markstory:
I'll do another release so the CakePHP pull request can be updated.

PR #67, opened 2015-12-07
Run appveyor on php 5.5.8

antograssiot:
This solution can't be applied directly to cakephp/cakephp as SQLServer driver are not included but it's still valid for Chronos.

I will look later in the week if there's a place where we could get the dll for cakephp and download them directly to the ext directory.


lorenzo:
Good job!

PR #68, opened 2016-01-01
Fixing HHVM

lorenzo:
* Fixed constructors signature, first argument default should have been 'now'

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -245,7 +245,7 @@ class ChronosInterval extends DateInterval
     public function __get($name)
     {
         switch ($name) {
-        case 'years':
+            case 'years':
                 return $this->isHHVM ? parent::__get('y') : $this->y;
 
             case 'months':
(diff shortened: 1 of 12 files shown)


markstory:
HHVM still doesn't like the code. I fixed a few more null values.

lorenzo:
I fixed the tests so they all pass locally in the newest hhvm. We need to wait for travis to upgrade their stack to see them passing there

src/ChronosInterval.php
```diff
@@ -238,30 +246,39 @@ public function __get($name)
...
+            case 'days':
+                return $this->isHHVM ? parent::__get('days') : $this->days;
+            case 'y':
+            case 'm':
+            case 'd':
+            case 'h':
+            case 'i':
+            case 's':
+            case 'invert':
+                return parent::__get($name);
```
markstory: Are these issues we should report to HHVM?

lorenzo: It is already reported, let me find the issue number and link it to this PR

lorenzo:
@markstory this one https://github.com/facebook/hhvm/issues/2952

Apparently it is fixed in the 3.11. I tested with 3.10

markstory:
@lorenzo Sounds like they've got it taken care of in a future release.

PR #70, opened 2016-01-04
Allow PHP 5.5.9

markstory:
This allows chronos to install on Ubuntu 14.04. Which is pretty important.

See cakephp/cakephp#7952


josegonzalez:
lolbuntu.

PR #72, opened 2016-01-05
Adding short explanation with main differences with Carbon

lorenzo:


README.md
```diff
@@ -28,8 +28,25 @@ use Cake\Chronos\Chronos;
 printf("Now: %s", Chronos::now());
 ```
 
+# Differences with nesbot/barbon
```
bcrowe: "carbon" typo

README.md
```diff
@@ -28,8 +28,25 @@ use Cake\Chronos\Chronos;
...
+The biggest and main difference is that `Chronos` extends `DateTimeImmutable` instead of `DateTime`.
+Immutability for date values has proven to be a great way of avoiding bugs and reduce the amount of code,
+since developers don't have to manually copy the instance every time they need a change.
+
+Another important feature it offers is the `Date` class, which is used for representing dates without time (calendar dates).
+Any time method called in this type of object is basically a no-op.
+
+A minor but still noticeable difference is that `Chronos` has no external dependencies, it is completely standalone.
+
+Finally, Chronos is faster than Carbon as it has been optimized for the creation of hundreds of intances with minimal
```
bcrowe: "instances" typo

README.md
```diff
@@ -28,8 +28,25 @@ use Cake\Chronos\Chronos;
...
 ```
 
+# Differences with nesbot/barbon
+
+The biggest and main difference is that `Chronos` extends `DateTimeImmutable` instead of `DateTime`.
+Immutability for date values has proven to be a great way of avoiding bugs and reduce the amount of code,
+since developers don't have to manually copy the instance every time they need a change.
+
+Another important feature it offers is the `Date` class, which is used for representing dates without time (calendar dates).
+Any time method called in this type of object is basically a no-op.
```
josegonzalez: `in` => `on`

README.md
```diff
@@ -28,8 +28,25 @@ use Cake\Chronos\Chronos;
...
+
+A minor but still noticeable difference is that `Chronos` has no external dependencies, it is completely standalone.
```
josegonzalez:
```
A minor - but still noticeable - difference is that `Chronos` has no external dependencies; Chronos is completely standalone.
```

README.md
```diff
@@ -28,8 +28,25 @@ use Cake\Chronos\Chronos;
...
+
+Finally, Chronos is faster than Carbon as it has been optimized for the creation of hundreds of intances with minimal
+overhead.
+
+Chronos also strives for HHVM compatibility, this library can be used safely with HHVM 3.11.
```
josegonzalez: We might want to point out that Carbon can still be used with PHP 5.3. We need 5.5.9 - actually 5.5.10 but Debian packaging, amirite - which *might* be a dealbreaker for users.

dereuromark:
Can someone with access also change the header descr `originally based off of carbon`? Carbon should be capital C as name.

markstory:
@dereuromark I updated the repo description.

Issue #71, opened 2016-01-04
RFC Datetime and Date class name


dereuromark:
The more I use it and think about it the more sense it would make to have a consistent naming scheme for the class names.
```php
//Currently
$time = new Chronos('2015-10-21 16:29:00');
$date = new Date('2015-10-21');
$time = new MutableDateTime('2015-10-21 16:29:00');
$date = new MutableDate('2015-10-21');

// Prosposed
$time = new DateTime('2015-10-21 16:29:00');
$date = new Date('2015-10-21');
$time = new MutableDateTime('2015-10-21 16:29:00');
$date = new MutableDate('2015-10-21');
```

Afaik the only reason was namespace clashing/confusion.
But with all the other 3 classes around I think the confusion is rather internal now, with Chronos not just as namespace but also as class name sticking out here.
`Chronos` also does not say anything about the type of date it contains or can work on ( is it Date, or DateTime or just Time or ...). For me that is a cool package name, but not a very explaning class name.
Thus I think **DateTime** would indeed be the more sane name for the Chronos class IMO.

We can change it now before releasing a 1.0.0.

lorenzo:
That could potentially be painful for end users, since using both \DateTime and Chronos\DateTime in the same class would require aliasing.

dereuromark:
And the other 3 don't have a core PHP counterpart you mean?

lorenzo:
Yeah, that's what I mean... In general I don't like class names to be repeated, even if we have namespaces, because reading the code becomes more difficult

dereuromark:
In that case pre- or suffixing is usually the way to go, but `ChronosDateTime` could become a little bit much to write maybe?

josegonzalez:
What about:

```php
class Chronos extends Chronos\DateTime
```

That way users that want to just use the Chronos version can, and all others can use the Chronos name.

dereuromark:
That sounds like a fair (and BC) compromise.

antograssiot:
PR #74 is open so discution can take place there.

PR #74, opened 2016-01-09
Introduce consistent DateTime class and use Chronos as alias.

dereuromark:
Resolves https://github.com/cakephp/chronos/issues/71
```
Cake\Chronos\Date;
Cake\Chronos\DateTime;
Cake\Chronos\MutableDate;
Cake\Chronos\MutableDateTime;

// Alias of DateTime
Cake\Chronos\Chronos;
```

The [README](https://github.com/cakephp/chronos/tree/master-date-time-class) points out the alias and that it is useful for migrating from Carbon, as it makes the process less likely to collide with stray core DateTime class usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,21 +23,21 @@ You can then use Chronos:
 <?php
 require 'vendor/autoload.php';
 
-use Cake\Chronos\Chronos;
+use Cake\Chronos\DateTime;
 
-printf("Now: %s", Chronos::now());
+printf("Now: %s", DateTime::now());
 ```
 
 # Differences with nesbot/carbon
 
-The biggest and main difference is that `Chronos` extends `DateTimeImmutable` instead of `DateTime`.
+The biggest and main difference is that `Cake\\Chronos\\DateTime` extends `DateTimeImmutable` instead of `DateTime`.
 Immutability for date values has proven to be a great way of avoiding bugs and reduce the amount of code,
 since developers don't have to manually copy the instance every time they need a change.
 
-Another important feature it offers is the `Date` class, which is used for representing dates without time (calendar dates).
+Another important feature it offers is the `Cake\\Chronos\\Date` class, which is used for representing dates without time (calendar dates).
 Any time method called on this type of object is basically a no-op.
 
-A minor but still noticeable difference is that `Chronos` has no external dependencies, it is completely standalone.
+A minor but still noticeable difference is that `Cake\\Chronos\\DateTime` has no external dependencies, it is completely standalone.
 
 Finally, Chronos is faster than Carbon as it has been optimized for the creation of hundreds of instances with minimal
 overhead.
@@ -45,8 +45,6 @@ overhead.
 Chronos also strives for HHVM compatibility, this library can be used safely with HHVM 3.11.
 
 # Migrating from Carbon
-
-
 First add `cakephp/chronos` to your `composer.json`:
 
 ```shell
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -23,21 +23,21 @@ You can then use Chronos:
 <?php
 require 'vendor/autoload.php';
 
-use Cake\Chronos\Chronos;
+use Cake\Chronos\DateTime;
 
-printf("Now: %s", Chronos::now());
+printf("Now: %s", DateTime::now());
 ```
```
markstory: Do you think it will be confusing that stdlib DateTime is mutabke, but this library uses the same name for an immutable object? It will be hard at a glance to know which Datetime a piece of code is using.

dereuromark: A fair concern.
Technically it's impossible to import both without aliasing one.
And if you use Chronos you would want to use it consistently across the code base.
Doing it half/half brings you only in trouble.

Apart from that, there should never be the use case of both in the same class file, unless upgrading partially from legacy code or alike. In those cases one can use the Chronos alias or use inline `\DateTime` maybe?

In general, nobody prevents you from the alias completely of course if that makes you feel more comfortable I guess.

src/Chronos.php
```diff
@@ -47,79 +50,6 @@
  * @property-read string  $timezoneName
  * @property-read string  $tzName
  */
-class Chronos extends DateTimeImmutable implements ChronosInterface
+class Chronos extends DateTime
```
josegonzalez: I think you extended the wrong class? This change is I *think* bc-incompatible?

dereuromark: No it is the right class. See the namespace above.

josegonzalez: Ah I see.

I hate that sort of automagic namespacing, but :+1:  :)

README.md
markstory: My concern is more that our DateTime class has completely different behavior from the DateTime class in stdlib.

dereuromark: At least our behavior is the better / more correct one ;)
As [this guy](https://www.reddit.com/r/PHP/comments/3zlvvb/chronos_a_standalone_datetime_library_originally/cynvr8b) put it.

Do you feel like this is a too big issue?

antograssiot: I do also find this confusing. I'm not a big fan of this idea personally :-1:

ADmad: At initial glance anyone reading code would be definitely confused. I too don't like this :-1:

src/Chronos.php
lorenzo: So we have the first confusion already :D

dereuromark:
We could keep Chronos as the documented one, only making DateTime available for those that really want to use it.

But if there is such a strong argument against that too, we can also close.

README.md
jippi: :-1:

jhli: Now let's add some real arguments. Chronos is a sexeh name whereas Datetime us not :troll:.

(I don't like a custom sexeh well made time lib masquerading as a php std lib)

ADmad:
> But if there is such a strong argument against that too, we can also close.

Every response has been either "no" or "this can be confusing / problematic". So better to just close.

markstory:
One thing we might want consider is aliasing `MutableDateTime` to `MutableChronos`. That would give the naming parity I think @dereuromark was going after without creating as much chance for confusion.

Issue #62, opened 2015-12-03
Find out if we can/need to workaround PHP bug


dereuromark:
See https://github.com/briannesbitt/Carbon/issues/539 and https://bugs.php.net/bug.php?id=63863

DateTime:setDate() date not used after modify("last day of...")

HavokInspiration:
Apparently there is a simple workaround given at the end of the PHP bug tracker ticket : https://bugs.php.net/bug.php?id=63863#1446180347

(Last comment).
Maybe it is worth trying it ?

dereuromark:
Wasnt there already some simililar issue - like in https://github.com/cakephp/chronos/pull/58/files ?

nicholas-eden:
I can verify that the workaround mentioned in that bug report works:
https://github.com/briannesbitt/Carbon/pull/542/files

markstory:
Pull request open now.

PR #80, opened 2016-02-06
Fix setDate() not working on some relative dates.

markstory:
When relative dates are created from last day of month, or first day of month expressions, the setDate() method does not work. This adds a workaround for a PHP issue. Sadly reflection was needed to access the `parent::setDate()` on an immutable object.

Refs #62


HavokInspiration:
Looks good to me.

PR #82, opened 2016-02-13
Fix createFromTimestamp() acting incorrectly.

markstory:
MutableDate was missing the important FrozenTimeTrait. Using this trait lets a pile of duplicate code to be removed. add(), sub() were made simpler by always calling setTime().

Refs #81


lorenzo:
CS tests are failing

PR #87, opened 2016-03-12
Removed redundant escape characters

Bilge:
There is no need to escape a hyphen `-` in this context. A hyphen is only a metacharacter in a character class (`[]`) and only when it does not appear as the first or last character in the class.


markstory:
Thanks.

PR #88, opened 2016-03-14
Fixed the create function when having a testNow.

lorenzo:
refs https://github.com/briannesbitt/Carbon/pull/633

### DIFF
--- a/src/Traits/FactoryTrait.php
+++ b/src/Traits/FactoryTrait.php
@@ -144,19 +144,31 @@ trait FactoryTrait
      */
     public static function create($year = null, $month = null, $day = null, $hour = null, $minute = null, $second = null, $tz = null)
     {
-        $year = ($year === null) ? date('Y') : $year;
-        $month = ($month === null) ? date('n') : $month;
-        $day = ($day === null) ? date('j') : $day;
+        $now = isset(static::$testNow) ? static::$testNow->getTimestamp() : time();
+        $defaults = array_combine(
+            [
+                'year',
+                'month',
+                'day',
+                'hour',
+                'minute',
+                'second',
+            ],
+            explode('-', date('Y-n-j-G-i-s', $now))
+        );
 
         if ($hour === null) {
-            $hour = date('G');
-            $minute = ($minute === null) ? date('i') : $minute;
-            $second = ($second === null) ? date('s') : $second;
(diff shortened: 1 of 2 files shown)


src/Traits/FactoryTrait.php
```diff
@@ -144,19 +144,31 @@ public static function minValue()
...
+        $defaults = array_combine(
+            [
+                'year',
+                'month',
+                'day',
+                'hour',
+                'minute',
+                'second',
+            ],
+            explode('-', date('Y-n-j-G-i-s', $now))
```
markstory: Do we need to take timezones into account here? Lower down this date is reconstructed with the `$tz` property, but this is missing a timezone.

lorenzo: I'll add a few more tests with time zones

lorenzo: @markstory What do you think it should happen? Do we create a date and assign the timezone without doing the conversion, or do you think the final result should alter the time if the timezone is different to the $testNow object?

markstory: Given that you are formatting a timestamp, could this use UTC internally and apply the timezone at the end? There is definitely tension between preserving `testNow` and honouring the parameters here.

lorenzo: Does applying the timezone in the end mean to alter the time from the testNow?

The way it works now is to take the time from testNow and assume it is from the passed timezone. That is, you get visually the same time, but it belongs to a different timezone. Thus, it is actually a different time.

markstory: But in the non-testNow cases, won't the time also be wrong if a timezone parameter is supplied? As all the defaults are coming from the default timezone.

lorenzo: That's my question... what of the two possible behaviors do you think is the one makes the most sense?

Do we convert to the provided TZ, or do we create the date, assign the TZ, but perform not transformation?

markstory: Not doing the transformation seems like it would be less desirable to me. My reasoning is, that when set TestNow to be a specific datetime that's the formatted datetime I'm expecting.

markstory:
@lorenzo Should this be merged now?

markstory:
Closing as this can't be merged anymore.

Issue #91, opened 2016-04-14
Relative string in Date constructor creates Date object with time not stripped


wreality:
Creating a new date object with a relative string in the constructor doesn't actually strip out the time component of the date object (but does correctly modify the date as otherwise expected) So something along the lines of:

```
$date = new Date("+7 days");
echo $date->format('Y-m-d H:i:s');
//Actual 2016-04-14 12:12:12
//Expected: 2016-04-14 00:00:00
```

Doing something like this is an OK work around:

```
$date = new Date();
$date = $date->modify('+7 days');
```

And, hoping its helpful, here's a failing testcase for it.

```
/**
 * @dataProvider dateClassProvider
 */
    public function testConstructWithRelative($class) {
        $c = new $class('+7 days');
        $this->assertEquals('00:00:00', $c->format('H:i:s'));
    }
```

markstory:
Thanks for the test case!

markstory:
Pull request open now.

PR #92, opened 2016-04-16
Fix Date constructor with relative strings.

markstory:
Check for relative strings when creating Date & MutableDate instances. Ensure that time values are removed from this constructor path.

Refs #91


HavokInspiration:
Looks good

Issue #44, opened 2015-10-28
ISO 8601: Missing colon in timezone


dereuromark:
Shouldn't the toIso8601String() method have an colon between the hours and minutes in the timezone part?
1975-12-25T14:15:16-0500 becomes 1975-12-25T14:15:16-05:00

Spec: http://www.w3.org/TR/NOTE-datetime

Refs https://github.com/briannesbitt/Carbon/issues/451

markstory:
We're using the [PHP constant](http://php.net/manual/en/class.datetime.php#datetime.constants.iso8601) The chances of that being wrong seem pretty low.

dereuromark:
There is a note below that linked documentation:

> Note: This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

dereuromark:
So either we fix it ourselves (add the colon) or should promote using the outlined formats.

ADmad:
> This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons.

Heh, so someone made a mistake initially and instead of fixing it they decided to stick with it. PHP is full of funnies.

markstory:
Bloody Hell. Well, we can use ATOM instead of ISO8601

markstory:
Pull request up.

PR #93, opened 2016-04-20
Fix toIso8601String(), which was previously returning ATOM strings.

davehensley:
toIso8601String() is returning ATOM strings, which are almost identical except for the colon in the TZ offset as documented [here](http://book.cakephp.org/3.0/en/chronos.html#formatting-strings) and [here](http://carbon.nesbot.com/docs/#api-commonformats).

### DIFF
--- a/src/Traits/FormattingTrait.php
+++ b/src/Traits/FormattingTrait.php
@@ -130,7 +130,7 @@ trait FormattingTrait
      */
     public function toIso8601String()
     {
-        return $this->format(DateTime::ATOM);
+        return $this->format(DateTime::ISO8601);
     }
 
     /**
--- a/tests/Date/StringsTest.php
+++ b/tests/Date/StringsTest.php
@@ -132,7 +132,7 @@ class StringsTest extends TestCase
     public function testToIso8601String($class)
     {
         $d = $class::create(1975, 12, 25, 14, 15, 16);
-        $this->assertSame('1975-12-25T00:00:00+00:00', $d->toIso8601String());
+        $this->assertSame('1975-12-25T00:00:00+0000', $d->toIso8601String());
     }
 
     /**
--- a/tests/DateTime/StringsTest.php
+++ b/tests/DateTime/StringsTest.php
@@ -144,7 +144,7 @@ class StringsTest extends TestCase
     public function testToIso8601String($class)
     {
         $d = $class::create(1975, 12, 25, 14, 15, 16);
-        $this->assertSame('1975-12-25T14:15:16-05:00', $d->toIso8601String());
+        $this->assertSame('1975-12-25T14:15:16-0500', $d->toIso8601String());
     }
 
     /**
--- a/tests/DateTime/TestingAidsTest.php
+++ b/tests/DateTime/TestingAidsTest.php
@@ -156,9 +156,9 @@ class TestingAidsTest extends TestCase
         $notNow = $class::parse('2013-07-01 12:00:00', 'America/New_York');
         $class::setTestNow($notNow);
 
-        $this->assertSame('2013-07-01T12:00:00-04:00', $class::parse('now')->toIso8601String());
-        $this->assertSame('2013-07-01T11:00:00-05:00', $class::parse('now', 'America/Mexico_City')->toIso8601String());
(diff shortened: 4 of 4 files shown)


dereuromark:
Wouldnt that just revert the fix https://github.com/cakephp/chronos/commit/685e8d37ec9dd1474a0a43b6e79ca2504ba5cd99 and therefore be actually a regression?

markstory:
The wikipedia article on [iso8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) has `2016-04-20T23:07:36+00:00` as the correct format. Which is what the code currently emits.

It looks like our documentation needs to be updated.

davehensley:
I see that you opened an [issue](https://github.com/cakephp/chronos/issues/44) (which was "resolved" by 685e8d37ec9dd1474a0a43b6e79ca2504ba5cd99) where you accuse PHP's internal DateTime::ISO8601 constant of being wrong. But it's not.

The link you provided in that issue documents the W3C format, which is a subset of ISO8601. From the [page you linked](http://www.w3.org/TR/NOTE-datetime):

> This profile defines a restricted range of formats, all of which are valid ISO 8601 dates and times.

So, all W3C dates are valid ISO 8601 dates, but not all ISO 8601 dates are valid W3C dates. So what you've really got there is the spec for toW3cString().

According to [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC), ISO 8601 does not require a colon in the timezone offset:

> The offset from UTC is appended to the time in the same way that 'Z' was above, in the form ±[hh]:[mm], ±[hh][mm], or ±[hh]. So if the time being described is one hour ahead of UTC (such as the time in Berlin during the winter), the zone designator would be "+01:00", "+0100", or simply "+01".

And this PR would at least make the method work as documented.

markstory:
I would rather change the documentation. Changing the code could break peoples applications.

davehensley:
The latest version of Carbon still uses PHP's constant, so leaving it as-is will break people's applications when they upgrade from CakePHP 3.1 to 3.2 (which is what happened to me) or if they're already familiar with Carbon or DateTime's perfectly valid implementations of toIso8601String(). Also, it's liable to confuse people (like me) who would expect this:

`$time->toIso8601String();`

to agree with this:

`$time->format($time::ISO8601);`

There are 4 methods which all output the exact same format, with the colon (toW3cString(), toRfc3339String(), toAtomString(), and now toIso8601String()). But now there's no method to directly get the colonless version anymore, so you need to use a workaround such as by passing the inherited ISO8601 constant into the inherited format() method like I did above.

Incidentally, CakePHP's DateFormatTrait::$_jsonEncodeFormat still uses the colonless version, which makes comparison unnecessarily difficult and confusing after using json_encode(), which is what's driving me to the above workaround.

Thanks for correcting the documentation.

ADmad:
@davehensley Quoting the PHP manual on [DateTime::ISO8601](http://php.net/manual/en/class.datetime.php#datetime.constants.iso8601) constant.

>     Note: This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

dakota:
@davehensley If you are so inclined as to actually read specs (I actually enjoy reading them) http://www.uai.cl/images/sitio/biblioteca/citas/ISO_8601_2004en.pdf

The ISO 8601 standard offers two different forms, the "basic" version, and the "extended" version. The basic version does not contain any separators except for `T` and `+/-` e.g. `19850412T101530+0400 `. The extended version does contain separators, e.g. `1985-04-12T10:15:30+04:00`. The standard (And the w3c subset) does not allow mixing of the two formats. This means that `1985-04-12T10:15:30+0400` is invalid.

> Incidentally, CakePHP's DateFormatTrait::$_jsonEncodeFormat still uses the colonless version, which makes comparison unnecessarily difficult and confusing after using json_encode(), which is what's driving me to the above workaround.

We need to get that fixed!

ADmad:
:+1: for fixing `DateFormatTrait::$_jsonEncodeFormat`

markstory:
I'll log a ticket in cakephp/cakephp.

PR #96, opened 2016-05-27
fixed doc of ’dayOfWeek’

iKenji:
refs [MagicPropertyTrait.php line44] (https://github.com/cakephp/chronos/blob/4e1dacba2e5e92283581c7fec195c652c9e99d09/src/Traits/MagicPropertyTrait.php#L44) 

|  |  |  |
|---|---|----|
| N  |  ISO-8601 numeric representation of the day of the week (added in PHP 5.1.0)  |   1 (for Monday) through 7 (for Sunday) |
> http://php.net/manual/en/function.date.php


dereuromark:
Thanks!

PR #103, opened 2016-08-11
Add toWeek() convenience method besides toQuarter().

dereuromark:
There is already toQuarter().
Usually toWeek() is also quite useful here as convenience wrapper.


markstory:
Looks like you have some extra whitespace still.

PR #107, opened 2016-11-11
Add PHP7.1 to test suite & fix compatibility.

markstory:
Refs #106

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
   - hhvm
 
 sudo: false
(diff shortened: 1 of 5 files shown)


lorenzo:
Is it expected for the build to fail?

dereuromark:
No, looks like there are more hidden bombs in the minor version jump regarding BC breaking.

markstory:
@lorenzo No I hoped it would pass with the 1 known problem fixed. But we didn't get lucky.

markstory:
I think I've gotten the tests fixed. I had to remove some tests as the addition of microtime in PHP7.1 has caused tests to no longer pass in some scenarios.

lorenzo:
Good job!

systematical84:
Has a fix for this been released on the 3.3 branch yet?

markstory:
Yes. You will need to use composer to update to the latest version of chronos

bravo-kernel:
✅

PR #108, opened 2016-11-12
Use codecov instead of coveralls.

ADmad:


dereuromark:
Sounds good to me!

markstory:
:+1:

Issue #114, opened 2016-12-21
Negative ChronosIntervals do not prepend a '-' in _toString()


rubenrangel:
Negative durations can exist. This PR uses the `invert` property from `DateInterval` to determine whether a negative sign should be prepended.

lorenzo:
Closing as there is a PR open

PR #115, opened 2016-12-21
Make ChronosInterval::toString() prepend '-' for negative intervals.

rubenrangel:
Fix for #114

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -461,6 +461,10 @@ class ChronosInterval extends DateInterval
             }
         }
 
-        return $specString === static::PERIOD_PREFIX ? 'PT0S' : $specString;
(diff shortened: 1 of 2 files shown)


src/ChronosInterval.php
```diff
@@ -461,6 +461,10 @@ public function __toString()
             }
         }
 
-        return $specString === static::PERIOD_PREFIX ? 'PT0S' : $specString;
+        if ($specString === static::PERIOD_PREFIX) {
+            return 'PT0S';
+        } else {
```
dereuromark: no else required.

rubenrangel: @dereuromark Should I make a new PR?

dereuromark: No need to. If you want to, you can.

PR #118, opened 2017-01-10
Do not specialize zero in DifferenceFormatter

serima:
Fixed that it was returned `1 second ago` even if there is no difference.
However, this modification affects the users of the library, so please consider it.

Thanks.

### DIFF
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -64,7 +64,7 @@ class Translator
      */
     public function plural($key, $count, array $vars = [])
     {
-        if ($count == 0 || $count == 1) {
+        if ($count === 0 || $count === 1) {
             return $this->singular($key, $vars);
         }
 
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -40,6 +40,7 @@ class Translator
         'from_now' => '{time} from now',
         'after' => '{time} after',
         'before' => '{time} before',
+        'just_now' => 'just now',
     ];
 
     /**
@@ -63,7 +64,7 @@ class Translator
      */
     public function plural($key, $count, array $vars = [])
     {
-        if ($count == 1) {
+        if ($count == 0 || $count == 1) {
             return $this->singular($key, $vars);
         }
 
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -58,7 +58,7 @@ class Translator
      * Get a plural message.
      *
      * @param string $key The key to use.
-     * @param string $count The number of items in the translation.
+     * @param int $count The number of items in the translation.
      * @param array $vars Additional context variables.
      * @return string The translated message or ''.
      */
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -40,7 +40,6 @@ class Translator
         'from_now' => '{time} from now',
         'after' => '{time} after',
         'before' => '{time} before',
-        'just_now' => 'just now',
     ];
 
     /**
@@ -64,7 +63,7 @@ class Translator
      */
     public function plural($key, $count, array $vars = [])
     {
-        if ($count === 0 || $count === 1) {
+        if ($count === 1) {
             return $this->singular($key, $vars);
         }
 
--- a/src/DifferenceFormatter.php
+++ b/src/DifferenceFormatter.php
@@ -77,11 +77,8 @@ class DifferenceFormatter
                 $unit = 'second';
                 break;
         }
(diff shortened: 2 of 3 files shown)


markstory:
I think this is a pretty reasonable change. It doesn't change any API contracts and just gives additional resolution in the `diffForHumans()` method.

src/Translator.php
```diff
@@ -63,7 +64,7 @@ public function exists($key)
      */
     public function plural($key, $count, array $vars = [])
     {
-        if ($count == 1) {
+        if ($count == 0 || $count == 1) {
```
dereuromark: Can't we make the count input var integer and have strict comparisons? All kind of things would say true here for comparison with == and 0 or 1 :)

serima: Thank you for reviewing :)
I changed loose comparisons to strict.

src/Translator.php
```diff
@@ -63,7 +64,7 @@ public function exists($key)
      */
     public function plural($key, $count, array $vars = [])
     {
-        if ($count == 1) {
+        if ($count === 0 || $count === 1) {
```
dereuromark: Not sure if you can just change the comparisons here, though.
See `@param string $count The number of items in the translation.` above,

serima: Ah, roger that! fixing.

serima: Updated phpdoc.
I think that properties of `ChronosInterval` are always integer. correct?

chinpei215:
I think this implementation is incorrect.
```php
$a = new \Cake\Chronos\Chronos('1999-01-23 01:23:45');
$b = new \Cake\Chronos\Chronos('1999-01-23 01:23:45');

debug( $a->diffForHumans($b) );
```
Why is this `just now`? It should be `0 seconds before` or something else. `just_now` should be used when `$isNow` equals to `true`. Also, I think we shouldn't change `$count == 1` to `$count == 0 || $count == 1`. Isn't "zero" plural in English?

markstory:
> Isn't "zero" plural in English?

It is.

src/Translator.php
markstory: @chinpei215 Brought up a good point. `0` is plural in english, which is what this translator implements. For example you would say 'I have zero shoes' not 'I have zero shoe'.

serima: @markstory @chinpei215 Thank you for reviewing. I agree with your points.

serima:
> It should be 0 seconds before or something else. just_now should be used when $isNow is true.

By the way what does @dereuromark think about this?
I completely agree with this point.

So I have just tried to fix it.
My approach : Do not specialize on zero. zero is just plural.

dereuromark:
@serima I didnt say you should blindly change it, we should investigate as to "how we can change it to make it stricter". It can also be a follow up PR to not mix this with your changes. But == 0 is bad since a lot of things will return true here.

serima:
@dereuromark 
> I didnt say you should blindly change it, we should investigate as to "how we can change it to make it stricter".

Yes, indeed. I'm sorry. 🙇

I have some questions.

> But == 0 is bad since a lot of things will return true here.

Where is `==0`? [here](https://github.com/cakephp/chronos/pull/118/commits/568d99328ea610c2a29744d4ba7f92610cbf7074#diff-cd35f60a79ff75867f0d5cbd2e9fbcf2R67)?

And should I continue `just_now` approach?

dereuromark:
Yeah it was only in your initial commit ( https://github.com/cakephp/chronos/pull/118/commits/568d99328ea610c2a29744d4ba7f92610cbf7074 ).
All good now.

Well, from my side just now makes still more sense than 1 sec ago for <= 0.
But let's see what others say.

chinpei215:
@serima In CakePHP3, `Cake\I18n\Time::timeAgoInWords()` would return `just now`, but `\Cake\I18n\Time::diffForHumans()` wouldn't return it. Actually, `just now` sounds weird to me if it is a result of diff. I think `same` would be better.

Apart from that, what if we add `timeAgoInWords()` method to Chronos? It would be better to separate pull request though.

serima:
@chinpei215 I really appreciate feedback in terms of CakePHP3. I think so too.
It certainly is an issue to be separated from this pull request.

chinpei215:
@serima OK. if you want, please open a new issue about adding `timeInAgoWords`. But it might be arguable as we have to change ChronosInterface..

PR #122, opened 2017-01-17
Add isNext- / isLast- methods

### LINKED ISSUE #119: Add isNext / isLast
Follow https://github.com/briannesbitt/Carbon/pull/846
to be more API upgrade consistent coming from Carbon to Chronos in the future.

serima:
Resolve #119

### DIFF
--- a/src/Traits/ComparisonTrait.php
+++ b/src/Traits/ComparisonTrait.php
@@ -246,7 +246,7 @@ trait ComparisonTrait
      */
     public function isNextWeek()
     {
-        return $this->toDateString() === static::now($this->tz)->addWeek()->toDateString();
+        return $this->weekOfYear === static::now($this->tz)->addWeek()->weekOfYear;
     }
 
     /**
@@ -256,7 +256,7 @@ trait ComparisonTrait
      */
     public function isLastWeek()
     {
-        return $this->toDateString() === static::now($this->tz)->subWeek()->toDateString();
+        return $this->weekOfYear === static::now($this->tz)->subWeek()->weekOfYear;
     }
 
     /**
@@ -266,7 +266,7 @@ trait ComparisonTrait
      */
     public function isNextMonth()
     {
-        return $this->toDateString() === static::now($this->tz)->addMonth()->toDateString();
+        return $this->month === static::now($this->tz)->addMonth()->month;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Traits/ComparisonTrait.php
```diff
@@ -240,6 +240,66 @@ public function isTomorrow()
     }
 
     /**
+     * Determines if the instance is within the next week
+     *
+     * @return bool
+     */
+    public function isNextWeek()
+    {
+        return $this->toDateString() === static::now($this->tz)->addWeek()->toDateString();
```
lorenzo: Would you mind changing this to avoid possible error in the future with timezones?

```php
static::now($this->tz) // Change to
$this->copy();
```

lorenzo: Also, this method seems incomplete or the description in the method is misleading. What does it mean to be "within" next week? This method seems to return true only if i is exactly next week.

serima: @lorenzo Thank you for your review!

> this method seems incomplete

Sorry. I re-checked them and fixed up. It should now work as described.

And I have a question about this.

```php
static::now($this->tz) // Change to
$this->copy();
```

https://github.com/briannesbitt/Carbon/pull/846
I think that every method is implemented to make a judgment based on "now" in this PR.

If we use `$this->copy()`, the behavior won't change?

src/Traits/ComparisonTrait.php
```diff
@@ -240,6 +240,66 @@ public function isTomorrow()
     }
 
     /**
+     * Determines if the instance is within the next week
+     *
+     * @return bool
+     */
+    public function isNextWeek()
+    {
+        return $this->weekOfYear === static::now($this->tz)->addWeek()->weekOfYear;
```
lorenzo: Can you change `static::now($this->tz)` to `$this->copy()` ?

serima: I can change it, but this test case will fail. What is the correct behavior? (please see also: https://github.com/cakephp/chronos/pull/122#discussion_r96583016)

```php
    public function testIsNextWeekTrue($class)
    {
        $this->assertTrue($class::now()->addWeek()->isNextWeek());
    }
```

lorenzo: Never mind, I did not think it through. Your code is correct

lorenzo:
thanks @serima !

PR #126, opened 2017-01-27
Update Date.php

DIDoS:
Remove overwriting of time zone parameter.
The parameter of the constructor is overwritten in the first line. This does not make sense.


markstory:
Uour changes [make tests fail](https://travis-ci.org/cakephp/chronos/jobs/195987647) in PHP5.x

markstory:
>  The parameter of the constructor is overwritten in the first line. This does not make sense.

It does if you take into account that `Date` objects don't have a time, and shouldn't have unique timezones either.

Issue #121, opened 2017-01-16
phpstan included


dereuromark:
I think similar to other projects including phpstan here would be really good in terms of code quality.

cc @ondrejmirtes

ondrejmirtes:
I'm all for that 😊 I can prepare a PR or supervise one.

markstory:
Sounds fine to me.

ondrejmirtes:
I opened a PR: #127

dereuromark:
Great, closing this issue then.

PR #127, opened 2017-01-31
Integrated PHPStan & related fixes

ondrejmirtes:
I ran [PHPStan](https://github.com/phpstan/phpstan) on your codebase and it found these issues. Feel free to comment on the fixes. I think it'd also to be beneficial to integrate PHPStan into your Travis build so you can prevent these errors in the future.

You can find out more about [PHPStan on Medium.com](https://medium.com/@ondrejmirtes/phpstan-2939cd0ad0e3).

This is response to #121 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,10 @@ matrix:
   include:
     - php: 7.0
       env: PHPCS=1 DEFAULT=0
+    - php: 7.0
+      env: PHPSTAN=1 DEFAULT=0
+    - php: 7.1
+      env: PHPSTAN=1 DEFAULT=0
 
   allow_failures:
     - php: hhvm
@@ -41,6 +45,7 @@ script:
   - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION != 7.0 ]]; then vendor/bin/phpunit; fi
 
   - if [[ $PHPCS = 1 ]]; then vendor/bin/phpcs -p --extensions=php --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests; fi
+  - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.6 && vendor/bin/phpstan analyse -c phpstan.neon -l 3 src tests; fi
 
 after_success:
   - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION = 7.0 ]]; then bash <(curl -s https://codecov.io/bash); fi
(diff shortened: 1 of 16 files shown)


ondrejmirtes:
I had to add this ignored error:

```
Call to an undefined static method Cake\\Chronos\\ChronosInterval::[a-zA-Z0-9_]+\(\)
```

Because PHPStan does not yet read `@method` annotations in a stable version.

ADmad:
CakePHP :heart: PHPStan

tests/Date/DateMutabilityConversionTest.php
```diff
@@ -10,10 +10,12 @@
  * @link          http://cakephp.org CakePHP(tm) Project
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
+namespace Cake\Chronos\Test\Date;
+
 use Cake\Chronos\Date;
 use Cake\Chronos\MutableDate;
 
-class DateMutabilityConversionTest extends TestCase
+class DateMutabilityConversionTest extends \TestCase
```
dereuromark: Should we have a use statement instead of inline FQCN?

ondrejmirtes: I updated the PR. But the build is failing and I suspect it's not because of my changes:

```
1) Cake\Chronos\Test\DateTime\DiffTest::testDiffForHumansRelative with data set #3 (Cake\Chronos\Chronos Object (...), Cake\Chronos\Chronos Object (...), '1 month before')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-1 month before
+4 weeks before
```

antograssiot: I bet it will work tomorrow on the first of February ;)

dereuromark: Maybe we should put the date to one a bit more away from the end of the month :)
Or fix it to be non-overflow.

.travis.yml
```diff
@@ -22,6 +22,10 @@ matrix:
   include:
     - php: 7.0
       env: PHPCS=1 DEFAULT=0
+    - php: 7.0
+      env: PHPSTAN=1 DEFAULT=0
+    - php: 7.1
+      env: PHPSTAN=1 DEFAULT=0
```
markstory: Why on both PHP 7.0 and 7.1?

ondrejmirtes: Because PHPStan uses the reflection of current PHP version it runs under, it can possibly find different issues on each one - for example in 7.1.1, third argument to `unpack` function was added so PHPStan can point out whether you are using the parameter on older versions.

markstory: Cool. Thanks for explaining.

ondrejmirtes:
This also failed because of PHPUnit.

antograssiot:
TestSylvain restarted and passing

dereuromark:
Really awesome work, thanks a lot!

Issue #129, opened 2017-02-10
Add __debugInfo()


robertpustulka:
Can we add `__debugInfo()` magic method for `Chronos`?

Especially `testNow` makes it hard to read chronos instances.

```
object(Cake\Chronos\Chronos) {
    date => '2017-02-10 14:22:51.195231'
    timezone_type => (int) 3
    timezone => 'UTC'
    [protected] toStringFormat => 'Y-m-d H:i:s'
    [protected] weekendDays => [
        (int) 0 => (int) 6,
        (int) 1 => (int) 7
    ]
    [protected] diffFormatter => null
    [protected] _lastErrors => []
    [protected] days => [
        (int) 1 => 'Monday',
        (int) 2 => 'Tuesday',
        (int) 3 => 'Wednesday',
        (int) 4 => 'Thursday',
        (int) 5 => 'Friday',
        (int) 6 => 'Saturday',
        (int) 7 => 'Sunday'
    ]
    [protected] weekStartsAt => (int) 1
    [protected] weekEndsAt => (int) 7
    [protected] relativePattern => '/this|next|last|tomorrow|yesterday|today|[+-]|first|last|ago/i'
    [protected] testNow => object(Cake\Chronos\Chronos) {
        date => '2017-02-10 14:22:51.195231'
        timezone_type => (int) 3
        timezone => 'UTC'
        [protected] toStringFormat => 'Y-m-d H:i:s'
        [protected] weekendDays => [
            (int) 0 => (int) 6,
            (int) 1 => (int) 7
        ]
        [protected] diffFormatter => null
        [protected] _lastErrors => []
        [protected] days => [
            (int) 1 => 'Monday',
            (int) 2 => 'Tuesday',
            (int) 3 => 'Wednesday',
            (int) 4 => 'Thursday',
            (int) 5 => 'Friday',
            (int) 6 => 'Saturday',
            (int) 7 => 'Sunday'
        ]
        [protected] weekStartsAt => (int) 1
        [protected] weekEndsAt => (int) 7
        [protected] relativePattern => '/this|next|last|tomorrow|yesterday|today|[+-]|first|last|ago/i'
        [protected] testNow => object(Cake\Chronos\Chronos) {
            //and so on and on...
        }
    }
}
```

lorenzo:
Yes, please... that would be great

robertpustulka:
@lorenzo How would this look like?

```
[...]

PR #131, opened 2017-02-13
Added __debugInfo()

robertpustulka:
See #129

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -127,4 +127,20 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
     {
         return $this;
     }
+
+    /**
+     * Return properties for debugging.
+     *
+     * @return array
+     */
+    public function __debugInfo()
+    {
+        $properties = [
+            'currentTime' => $this->format('Y-m-d H:i:s.u'),
+            'timezone' => $this->getTimezone()->getName(),
+            'hasFixedNow' => isset(self::$testNow)
+        ];
+
+        return $properties;
+    }
 }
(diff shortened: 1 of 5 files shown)


lorenzo:
Would you mind adding a couple simple tests? Only to prevent bad method calls in the future

src/Chronos.php
```diff
@@ -127,4 +127,20 @@ public function copy()
...
+
+    /**
+     * Return properties for debugging.
+     *
+     * @return array
+     */
+    public function __debugInfo()
+    {
+        $properties = [
+            'currentTime' => $this->format('Y-m-d H:i:s.u'),
```
markstory: currentTime implies 'now' and not this object's time.

dereuromark: Would `value` work?

robertpustulka: Maybe just `time` and `date`?

inoas: `date`, `time`, `timezone`?

markstory: time and date work for me.

lorenzo:
Thanks!

src/MutableDateTime.php
```diff
@@ -175,7 +175,7 @@ public function __set($name, $value)
     public function __debugInfo()
     {
         $properties = [
-            'currentTime' => $this->format('Y-m-d H:i:s.u'),
+            'time' => $this->format('Y-m-d H:i:s.u'),
```
dereuromark: dateTime

Issue #133, opened 2017-03-31
midnight is not recognized as a relative keyword


gzumba:
Midnight is used by Chronos::today() and as it is not recognized to be a relative keyword, tests using ::today() will not work correctly with the setTestNow()

markstory:
Closing as a pull request is open.

PR #134, opened 2017-03-31
Add midnight to the list of relative keywords.

gzumba:
Midnight is used by Chronos::today() which will now return the correct
testNow value. Fixes issue #133


markstory:
Thanks!

Issue #135, opened 2017-04-01
DST handling while date modifying


Ovsyanka:
Hello!

What do you think about implement the same DST handling like i suggested in [issue](https://github.com/briannesbitt/Carbon/issues/883) for Carbon?

For example i have the date `2014-03-30 00:00:00` in `Europe/London` (`new Carbon('2014-03-30 00:00:00, 'Europe/London')`), then if i want to increase date by 1 day, i expect `2014-03-31 00:00:00`, but if want to increase date by 24 hours, then i expect `2014-03-31 01:00:00`, because in this timezone there will be that time after 24 hours (timezone offset changes because of Daylight correction). The same for minutes and seconds.

As i see, in chronos:

```php
$date = new Chronos('2014-03-30 00:00:00', 'Europe/London');
$date = $date->modify('+86400 sec');
$this->assertSame('2014-03-31 01:00:00 +01:00', $date->format('Y-m-d H:i:s P'));

// failed:
// -2014-03-31 01:00:00 +01:00
// +2014-03-31 00:00:00 +01:00
```

If you agree with me i will write pull-request.

markstory:
I think what you've outlined makes sense to me. Advancing by hours, minutes or seconds and crossing the DST boundary should take the DST transition into account, as the transition happened, and a 'bonus' hour was gained/lost.

Ovsyanka:
Okay, i will prepare pull-request in a few days.

Ovsyanka:
I have new thoughts about the issue. Please look to issue in Carbon project: https://github.com/briannesbitt/Carbon/issues/883#issuecomment-339446962

markstory:
It sounds like you're proposing to effectively revert the previous set of changes. I may have misunderstood the discussion thread in Carbon though.

Ovsyanka:
Not exactly revert. I am not sure about decision we standed to change how `modify` method works itself. Instead I suggeted to move that changes particulary in `add|sub` methods of `Carbon` (and `Chronos`), only the `sec|minute|hour` ones.

In addition I [realised](https://github.com/briannesbitt/Carbon/pull/884#discussion_r145051548), that my current code isn't perfect right now and it runs when `ordinal unit` `second` used, not only `time unit` `second`.

Besides that I investigate independent bug, that with negative time changing to skipped time period (between 1:00 and 2:00) the time changing to +1 hour instead -1 hour.

itsjavi:
[...]

PR #138, opened 2017-04-03
Implementation of issue #135 DST handling while date modifying

Ovsyanka:
While writing this code i realised that i dont like this "Traits" approach. And i think it will be better to implement two different overloades of modify method for both mutable and immutable objects. But i am not sure, so i asking yfor your suggestions and corrections. Thank you!

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -129,6 +129,37 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
     }
 
     /**
+     * Overloading original modify method to handling modification with DST change
+     *
+     * For example, i have the date 2014-03-30 00:00:00 in Europe/London (new Carbon('2014-03-30 00:00:00,
+     *   'Europe/London')), then if i want to increase date by 1 day, i expect 2014-03-31 00:00:00, but if want to
+     *   increase date by 24 hours, then i expect 2014-03-31 01:00:00, because in this timezone there will be that time
+     *   after 24 hours (timezone offset changes because of Daylight correction). The same for minutes and seconds.
+     *
+     * @param string $modify argument for php DateTime::modify method
+     *
+     * @return static
+     */
+    public function modify($modify)
+    {
+        if (!preg_match('/(sec|second|min|minute|hour)s?/i', $modify)) {
+            return parent::modify($modify);
+        }
+
+        $timezone = $this->getTimezone();
+
+        // I use parent modify method only if the current object timezone is UTC
+        // If it is not - i making another object in UTC wich will call to parent modify
+        if ($timezone->getName() !== 'UTC') {
+            $date = $this->setTimezone('UTC')->modify($modify)->setTimezone($timezone);
+        } else {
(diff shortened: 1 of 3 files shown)


src/Traits/ModifyExtendTrait.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        $timezone = $this->getTimezone();
+        $timezoneName = $timezone->getName();
+
+        // Here is tricky code for immutable DateTime
+        // I use parent modify method only if current object timezone is UTC
+        // If it is not - i making another object in UTC wich will call to parent modify
+        if ($timezoneName !== 'UTC') {
+            $date = $this->setTimezone('UTC');
```
ADmad: You can used chained method calls instead of assigning result of each call to variable.

Ovsyanka: Yeah, i know. I think that my way is more clear. But it is quite "matter of opinion" question, so I will chain it in that case because there only three operations.
Thank you.

src/Traits/ModifyExtendTrait.php
```diff
@@ -0,0 +1,54 @@
...
+        if (!preg_match('/(sec|second|min|minute|hour)s?/i', $modify)) {
+            return parent::modify($modify);
+        }
+
+        $timezone = $this->getTimezone();
+        $timezoneName = $timezone->getName();
+
+        // Here is tricky code for immutable DateTime
+        // I use parent modify method only if current object timezone is UTC
+        // If it is not - i making another object in UTC wich will call to parent modify
```
lorenzo: If you need this branch for the immutable classes, then it is better to put the method there directly or have separate traits for each.

Ovsyanka: Actually this code works for both types, but its "tricky" structure needed mainly for immutable type.
I implemented this stuff in one trait because i see that approach (using traits with universal code) in existed code. But i asked in my pull-request message what do you think about this - should i leave it as is or should i implement two versions of that code for each type of  DateTime objects. I think it is better to talk about it according to whole pull-request and not only in "code comments"

lorenzo:
@Ovsyanka I would prefer the methods to be in different classes instead of the trait

Ovsyanka:
Okay, so we have two voices (mine and yours, @lorenzo). I will make new version with separate implementations.

markstory:
There are a bunch of PHPStan errors but they don't appear to be related to this. I can fix them separately.

Ovsyanka:
It is great, because i am a little confused with that errors. Where you will fix it? In the master branch? I just have to wait and rebase?

markstory:
@Ovsyanka I can fix those errors in a separate pull request. We can re-run your builds once hat

ADmad:
I too don't get any errors locally running phpstan at level 4 locally for `src` and `tests` folder.

markstory:
Going to merge in, and we can skip the errors on `master` if they continue.

PR #139, opened 2017-04-04
When i was doing pull-request i has problems vith validating code style. I has to look at travis jobs to see how to validate it properly.

Ovsyanka:
So i suggest to add this phpcs config and composer scripts to make it simpler to newcomers.
Besides that now i dont need to search (if i forgot it) and write the whole comman: `vendor\bin\phpcs -p --extensions=php --standard=vendor\cakephp\cakephp-codesniffer\CakePHP .\tests .\src`, i can just write `composer phpcs` or even `phpcs` if i have phpcs installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,5 +44,14 @@
       "Cake\\Chronos\\Test\\": "tests"
     },
     "files": ["tests/TestCase.php"]
+  },
+  "scripts": {
+    "check": [
+      "@test",
+      "@cs-check"
+    ],
+    "test": "phpunit",
+    "cs-check": "phpcs",
+    "cs-fix": "phpcbf"
   }
 }
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,14 @@
+<?xml version="1.0"?>
+<ruleset name="PHP_CodeSniffer">
+ <description>Default configuration</description>
+
+ <rule ref="vendor/cakephp/cakephp-codesniffer/CakePHP"/>
+
+ <file>./src</file>
+ <file>./tests</file>
+
+ <arg phpcs-only="true" value="p"/>
+ <arg name="extensions" value="php"/>
+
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -44,5 +44,10 @@
       "Cake\\Chronos\\Test\\": "tests"
     },
     "files": ["tests/TestCase.php"]
+  },
+  "scripts": {
+      "phpunit": "phpunit",
+      "phpcs": "phpcs",
```
dereuromark: those should be the standard keys: "cs-check" and "cs-fix".
phpunit should be "test". See cakephp/app repo.

Ovsyanka: Thank you! Fixed.

phpcs.xml.dist
```diff
@@ -0,0 +1,14 @@
...
+ <rule ref="vendor/cakephp/cakephp-codesniffer/CakePHP"/>
+
+ <file>./src</file>
+ <file>./tests</file>
+
+ <arg phpcs-only="true" value="p"/>
+ <arg name="extensions" value="php"/>
+
+ <config name="colors" value="1"/>
+</ruleset>
```
dereuromark: new line missing here.

dereuromark:
:+1:

phpcs.xml.dist
Ovsyanka: Do you mean before `</ruleset`>?

Ovsyanka: Is there any place where described that rules for xml? It will be nice to have automatic xml style checking too.

Ovsyanka: Ah, okay. Fixed.
Didn't know that it is matter to xml files too. Now i searched about that and founded the [explanation](http://stackoverflow.com/a/729795/3653002) why.

markstory:
Thanks!

PR #142, opened 2017-04-17
Few phpstan related fixes

Ovsyanka:
First of all i added phpstan in composer dev dependencies, because it running on travis and developers should be able to run it before pushing pull-requests.
In addition i added composer script "phpstan" and included it in "check" script.

Some ignoring errors in phpstan is actually right. And code shold be fixed.
Some appearing errors is wrong and should be ignored.
All the stuff you can see in the commits.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,5 @@
 parameters:
 	ignoreErrors:
-		- '#Call to an undefined method Cake\\Chronos\\ChronosInterface::now\(\)#'
-
 		# Appears because interface can't describe properties and formally it doesen't have properties.
 		- '#Call to an undefined static method Cake\\Chronos\\ChronosInterval::[a-zA-Z0-9_]+\(\)#'
 		- '#Access to an undefined property Cake\\Chronos\\ChronosInterface::\$tz#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,9 @@
 parameters:
 	ignoreErrors:
-		- '#Call to an undefined static method Cake\\Chronos\\ChronosInterval::[a-zA-Z0-9_]+\(\)#'
 		- '#Call to an undefined method Cake\\Chronos\\ChronosInterface::now\(\)#'
+
+		# Appears because interface can't describe properties and formally it doesen't have properties.
+		- '#Call to an undefined static method Cake\\Chronos\\ChronosInterval::[a-zA-Z0-9_]+\(\)#'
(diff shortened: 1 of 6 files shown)


src/Traits/ComparisonTrait.php
```diff
@@ -144,7 +144,7 @@ public function between(ChronosInterface $dt1, ChronosInterface $dt2, $equal = t
      *
      * @param \Cake\Chronos\ChronosInterface $dt1 The instance to compare with.
      * @param \Cake\Chronos\ChronosInterface $dt2 The instance to compare with.
-     * @return static
+     * @return ChronosInterface
```
dereuromark: Should be FQCN then.

Ovsyanka: I am not against it, but why? Can you point me to some explanation?

cleptric: IDE autocompletion.  Also, it's a convention we follow through the framework :)

Ovsyanka: Okay, Fixed it

Ovsyanka:
I see the problem with travis. I plat to exclude phpstan from dependencies if php version below 7. If you have any other suggestions - please say it!

cleptric:
@Ovsyanka Have a look at the `.travis.yml` over at the [cakephp repo](https://github.com/cakephp/cakephp/blob/master/.travis.yml)

dereuromark:
Can we then please also follow https://github.com/cakephp/cakephp/pull/10520 once it is merged?

Ovsyanka:
@dereuromark, yes, I can. But I want to say my opinion about that.

I think, that it is not good to store such things it the tests folder because there expected code tests. I mean tests, that actually runs code. In our case it is phpunit tests. Phpstun is more like code style checking.

I dont think that configuration files in the root folder is bad. Because it is standard place for it. And you can put it in some different folder instead tests, like "configs", "style-checking", etc.

But it is only my opinion and you may count it or not. By now I will do as you suggested.

phpstan.neon
```diff
@@ -1,5 +1,11 @@
 parameters:
 	ignoreErrors:
+		# Apears because interface can't describe properties and formally it doesen't have properties.
```
markstory: Typo in 'Appears'

Ovsyanka:
Okay, It looks good now, I only have to to clear up commits history. Unfortunately, I can do it only tomorrow, but you have time to give me mooore crititism =)

Ovsyanka:
It is done.

Issue #143, opened 2017-04-24
Add @property comment for magic properties


xavier83ar:
It would be great to add `@property` comments to `Cake\Chronos\Traits\MagicPropertyTrait` for magic properties, this way we get hinting support for them and avoid "accessed via magic method" soft warning on some IDE's.

dereuromark:
Are you able to make a PR with these changes? That would be great!

xavier83ar:
Ok.

PR #144, opened 2017-04-24
Magic property annotations

xavier83ar:
https://github.com/cakephp/chronos/issues/143

### DIFF
--- a/src/Traits/MagicPropertyTrait.php
+++ b/src/Traits/MagicPropertyTrait.php
@@ -19,6 +19,32 @@ use InvalidArgumentException;
 /**
  * Provides the magic methods that allow read access
  * to magic properties.
+ *
+ * @property int year
+ * @property int yearIso
+ * @property int month
+ * @property int day
+ * @property int hour
+ * @property int minute
+ * @property int second
+ * @property int micro
+ * @property int dayOfWeek
+ * @property int dayOfYear
+ * @property int weekOfYear
+ * @property int daysInMonth
+ * @property int timestamp
+ * @property int weekOfMonth
+ * @property int age
+ * @property int quarter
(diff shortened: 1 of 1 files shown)


src/Traits/MagicPropertyTrait.php
```diff
@@ -19,6 +19,32 @@
 /**
  * Provides the magic methods that allow read access
  * to magic properties.
+ *
+ * @property-read int year
```
dereuromark: Arent those usually variables? $year etc?
See the other annotations in other files.

xavier83ar: Mm... yeah, now I see that those annotations exists on Cake\Chronos\Chronos and Cake\Chronos\Date as well. Now I see that I was using an outdated version of Chronos (1.0.0), my bad, which doesn't have those annotations. However, isn't the trait a better place for them? As it's the trait which provides that properties...

markstory: They are better on the trait, but could the annotations be made consistent?

xavier83ar: What do you mean?

markstory: I agree that the magic property annotations make sense to have on the traits that provide them. However, it would be better if the added properties used the same conventions used in the rest of the library.

xavier83ar: Oh, I understand, and I agree with you even if I don't actually know what conventions do you refer to, but if it's about homogenize code and making it consistent, I'll alway agree. First thing I came in mind, seeing the recent changes on the framework, would be to have getters methods (may be magic ones) instead of properties, but let me know what's your idea.

PR #146, opened 2017-06-16
Limited PHP version constraint upper bound

Bilge:
We cannot possibly know we're compatible with versions of PHP never before created, so we should not claim to be.

For example, PHP 7.1 support was added in [version `1.0.1`](https://github.com/cakephp/chronos/releases/tag/1.0.1), but users of earlier versions don't know they need to upgrade because earlier versions claim to be compatible with `"php": ">=5.5.9"`, meaning version 5.5.9 *and every future version of PHP*, which is strictly incorrect unless you possess profound time travelling capabilities.

This PR only seeks to claim compatibility with future versions of the same major version of PHP as we have tested compatibility for, but even that is a bit risky. It may be better to specify `<7.2` instead of `^7` and enforce the rule that we **do not claim to be compatible with any version of PHP we have not tested against**, instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "source": "https://github.com/cakephp/chronos"
   },
   "require": {
-    "php": ">=5.5.9"
+    "php": "^5.5.9|^7"
   },
   "require-dev": {
     "phpunit/phpunit": "<6.0",


composer.json
```diff
@@ -26,7 +26,7 @@
     "source": "https://github.com/cakephp/chronos"
   },
   "require": {
-    "php": ">=5.5.9"
+    "php": "^5.5.9|^7"
```
prisis: You should use a double pipe

https://getcomposer.org/doc/articles/versions.md#writing-basic-version-constraints

prisis: There is no benefit, composer supports both

dereuromark:
We usually expect current compatability until proven otherwise, e.g. via Travis. This is done gar earlier than production use, has worked so far.
:-1:

amcsi:
@Bilge The `^7` bit should probably be `^7.1`.

Had originally before the 7.1 fixes composer.json had `>=5.5.9 | ^7` in it originally while a project was running under PHP 7.0, that project would have suffered the same problems once moving to PHP 7.1, because ^7 would allow for 7.1 as well.

josegonzalez:
@Bilge the whole point of chronos is to handle time-manipulation, so I'm guessing its safe to say we do/

markstory:
Merging as the current state will only be a problem when PHP 8 comes out.

Bilge:
Thanks @markstory. I appreciate you exercising your judgement to merge this despite strong opposition from commentators.

PR #156, opened 2017-12-25
Fix isNextWeek(), isLastWeek(), isNextMonth() and isLastMonth()

chinpei215:
Currently these methods ignore the year. For example, if poeple do:
```php
(new Chronos('2000-01-01'))->isNextMonth();
```
on 2017-12-01, it will return true unexpectedly.


ADmad:
:clap:

lorenzo:
Thanks!

PR #158, opened 2018-01-13
Optimize the magic properties getter

vlakoff:
I think the optimization is worth it, as this method may be called thousands of times.

* Use the language construct `isset()` instead of the function `array_key_exists()`
* Use a static variable
----
(resubmit of https://github.com/briannesbitt/Carbon/pull/1017 which was targetted at Carbon)


lorenzo:
Nice, thanks so much for this optimization!

PR #159, opened 2018-01-17
reverted issue #135 DST handling while date modifying

Ovsyanka:
The approach was wrong. Because not all users expect the same behavior of modify in matter of handling DST changes. There is use cases for both approaches in both diapasones of units (less than day and equal or bigger). And by default php already have the way to handle this. You just should

* use UTC timezone if you don't want to have timestamp adjusments in case of DST changes
* use non-UTC timezone if you want to have such adjustments


chinpei215:
When I saw the modify method first time, it seemed weird to me too. But I am not sure we should revert it as existing applications would rely on the behavior.
Instead, I would suggest you to call the parent method directly like the following:
```php
call_user_func([$time, 'parent:: modify'], '+1 hour');
```

Ovsyanka:
Which modify method was weird for you? DateTime::modify or Chronos::modify?

I don't think that this behavior widely used because it was implemented not a long time  ago. And I see that more like an error now. At least because you can't use default PHP handling at this moment.

And I don't understand your suggestion. Where and why we should use code you suggested?

Ovsyanka:
@markstory Seems like the falling phpstan tests does not related to my changes. Does they?
```
Ignored error pattern #Property Cake\\Chronos\\MutableDateTime::\$(tz|timezone) \(DateTimeZone\) does not accept string# was not matched in reported errors.  
  Ignored error pattern #Access to an undefined property Cake\\Chronos\\MutableDateTime::\$doesNotExit# was not matched in reported errors.                     
```

chinpei215:
@Ovsyanka 
> Which modify method was weird for you? DateTime::modify or Chronos::modify?

Chronos::modify(). Overriding for bugfix is a wrong approach.

> I don't think that this behavior widely used because it was implemented not a long time ago. And I see that more like an error now. 

I didn't know you added Chronos::modify() 10 months ago. I thought it came from Carbon.  Then I agree to remove the modify method.

> At least because you can't use default PHP handling at this moment.

No, you can use default PHP handling if you want. That is the way I suggested above:
```php
$time = new Chronos('2014-03-30 00:00:00', 'Europe/London');
debug( $time->modify('+24 hours') ); // 2014-03-31 01:00
debug( call_user_func([$time, 'parent::modify'], '+24 hours') ); // 2014-03-31 00:00:00
```

> And I don't understand your suggestion. Where and why we should use code you suggested?

You can use the above code to call DateTime::modify() via a Chronos instance at any place. But, as I said above, I am OK with reverting the changes.

Ovsyanka:
Thank you for such detailed answer =) I got it now. Glad that you agree with me at the end.

PR #162, opened 2018-02-26
Add addYearsWithOverflow()

### LINKED ISSUE #157: addYear and leap years
When I call `addYear` on a 29th of February I'd expect to stay in the same month and have the 28th of February in the next year, however the result is the first of March.

```php
// Expected: 2013-02-28 12:00:00.000000
var_dump((new \DateTime('2012-02-29 12:00:00'))->modify('+1 year')); // 2013-03-01 12:00:00.000000
var_dump((new Chronos('2012-02-29 12:00:00'))->addYear(1)); // 2013-03-01 12:00:00.000000
```

Chronos got the same behavior when using addMonth:
```php
var_dump((new \DateTime('2014-05-31 12:00:00'))->modify('+1 month')); // 2014-07-01 12:00:00.000000
var_dump((new Chronos('2014-05-31 12:00:00'))->addMonth(1)); // 2014-06-30 12:00:00.000000
```

I can provide a pull request if there aren't any objections to changing this.

serima:
Resolve #157 by @chinpei215's way.

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -512,6 +512,17 @@ interface ChronosInterface extends DateTimeInterface
      * Add years to the instance. Positive $value travel forward while
      * negative $value travel into the past.
      *
+     * If the new date does not exist, the last day of the month is used
+     * instead instead of overflowing into the next month.
+     *
+     * ### Example:
+     *
+     * ```
+     *  (new Chronos('2015-01-03'))->addYears(1); // Results in 2016-01-03
+     *
+     *  (new Chronos('2012-02-29'))->addYears(1); // Results in 2013-02-28
+     * ```
+     *
      * @param int $value The number of years to add.
      * @return static
      */
@@ -520,30 +531,44 @@ interface ChronosInterface extends DateTimeInterface
     /**
      * Add a year to the instance
      *
+     * Has the same behavior as `addYears()`.
+     *
      * @param int $value The number of years to add.
      * @return static
      */
     public function addYear(int $value = 1): self;
 
     /**
-     * Remove a year from the instance
-     *
-     * @param int $value The number of years to remove.
-     * @return static
-     */
-    public function subYear(int $value = 1): self;
-
-    /**
      * Remove years from the instance.
      *
+     * Has the same behavior as `addYears()`.
+     *
      * @param int $value The number of years to remove.
      * @return static
      */
-    public function subYears(int $value): self;
+    public function subYears($value);
+
+    /**
+     * Remove a year from the instance.
+     *
+     * Has the same behavior as `addYears()`.
+     *
+     * @param int $value The number of years to remove.
+     * @return static
+     */
+    public function subYear($value = 1);
 
     /**
      * Add years with overflowing to the instance. Positive $value
      * travels forward while negative $value travels into the past.
+     *
+     * If the new date does not exist, the days overflow into the next month.
+     *
+     * ### Example:
+     *
+     * ```
+     *  (new Chronos('2012-02-29'))->addYearsWithOverflow(1); // Results in 2013-03-01
+     * ```
      *
      * @param int $value The number of years to add.
      * @return static
@@ -553,21 +578,17 @@ interface ChronosInterface extends DateTimeInterface
     /**
      * Add a year with overflow to the instance
      *
+     * Has the same behavior as `addYearsWithOverflow()`.
+     *
      * @param int $value The number of years to add.
      * @return static
      */
     public function addYearWithOverflow($value = 1);
 
     /**
-     * Remove a year with overflow from the instance
-     *
-     * @param int $value The number of years to remove.
-     * @return static
-     */
-    public function subYearWithOverflow($value = 1);
-
-    /**
      * Remove years with overflow from the instance
+     *
+     * Has the same behavior as `addYearsWithOverflow()`.
      *
      * @param int $value The number of years to remove.
      * @return static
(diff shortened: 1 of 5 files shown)


src/Traits/ModifierTrait.php
```diff
@@ -243,12 +243,31 @@ public function second($value)
...
      */
     public function addYears($value)
     {
-        return $this->modify((int)$value . ' year');
+        $month = $this->month;
+        $date = $this->modify((int)$value . ' year');
+
+        if ($date->month !== $month) {
+            return $date->modify('last day of previous month');
+        }
```
markstory: This changes the semantics of `addYears()` in a good way, but this and the interface changes will require a minor release bump.

src/ChronosInterface.php
```diff
@@ -539,6 +539,39 @@ public function subYear($value = 1);
...
     public function subYears($value);
 
+    /**
+     * Add years with overflowing to the instance. Positive $value
+     * travels forward while negative $value travels into the past.
+     *
+     * @param int $value The number of years to add.
+     * @return static
+     */
+    public function addYearsWithOverflow($value);
```
dereuromark: Do we have to major for these interface methods?

markstory:
The PHPStan errors look like they were caused by changes in PHPStan I'll get those fixed up.

saeideng:
I'm :-1:   with this changes 
this can be  make change behavior on the other calendar that rely on Chronos

saeideng:
@dereuromark   I tested just with `2012-02-29`
```php
    $time1=new \Cake\I18n\Time('2012-02-29');
    $time2=(new \Cake\I18n\Time('2012-02-29'))->addYears(1);
    $time3=new \Cake\I18n\Time('2013-02-28');

    debug($time1);
    debug($time2);
    debug($time3);

    debug($time1->i18nFormat('yyyy-MM-dd', null, 'en-IR@calendar=persian'));
    debug($time2->i18nFormat('yyyy-MM-dd', null, 'en-IR@calendar=persian'));
    debug($time3->i18nFormat('yyyy-MM-dd', null, 'en-IR@calendar=persian'));

    //2012-02-29
    //2013-03-01
    //2013-02-28

    // 1390-12-10
    // 1391-12-11
    // 1391-12-10
```

saeideng:
>and added only in next major

:+1:

dereuromark:
So will this go into next major and 2.x branch?

markstory:
@dereuromark Yes, I think it needs a rebase.

src/Traits/ModifierTrait.php
```diff
@@ -243,12 +243,31 @@ public function second($value)
      * Add years to the instance. Positive $value travel forward while
      * negative $value travel into the past.
      *
+     * When adding or subtracting years, if the resulting time is a date
+     * that does not exist, the result of this operation will always be the
+     * last day of the intended month.
+     *
```
othercorey: I would leave a suggestion, but I can't replace multi-line comments.

Suggestion:
If the new date does not exist, the last day of the month is used instead instead of overflowing into the next month.

src/Traits/ModifierTrait.php
```diff
@@ -243,12 +243,31 @@ public function second($value)
...
+     * ### Example:
```
othercorey: This docblock should be on the interface.

src/ChronosInterface.php
```diff
@@ -539,6 +539,39 @@ public function subYear($value = 1);
...
+
+    /**
+     * Add a year with overflow to the instance
+     *
```
othercorey: This is missing an explanation of overflow and needs the same example as ``addYears()`` but with the new result.

If the new date does not exist, the date overflows into the next month.

src/ChronosInterface.php
```diff
@@ -539,6 +539,39 @@ public function subYear($value = 1);
      */
     public function subYears($value);
 
+    /**
+     * Add years with overflowing to the instance. Positive $value
+     * travels forward while negative $value travels into the past.
+     *
```
othercorey: This is missing an explanation of overflow and needs the same example as ``addYears()`` but with the new result.

If the new date does not exist, the date overflows into the next month.

src/Traits/ModifierTrait.php
```diff
@@ -284,6 +303,51 @@ public function subYears($value)
...
+
+    /**
+     * Add a year with overflow to the instance
+     *
+     * @param int $value The number of years to add.
+     * @return static
+     */
+    public function addYearWithOverflow($value = 1)
+    {
+        return $this->modify((int)$value . ' year');
```
othercorey: This can just call addYearsWithOverflow($value) as it's an alias.

tests/DateTime/AddTest.php
```diff
@@ -374,6 +374,48 @@ public function testAddYearPassingArg($class)
         $this->assertSame(1977, $class::createFromDate(1975)->addYear(2)->year);
     }
 
+    /**
+     * @dataProvider classNameProvider
+     * @return void
+     */
+    public function testAddYearWithOverflow($class)
```
othercorey: These tests show go next to the existing addYears() test in the file.

othercorey:
I cleaned up the PR and re-based with 2.x.

There is some noise where I re-ordered the interface to be consistent.  addYears() then addYear(), etc.

I added examples to overflow, aligned the interface and trait docblock and simply referenced the addYears or addYearsWithOverflow in each of the specialized functions instead of maintaining 20 different full explanations.

markstory:
Seems like a good fit for the 2.x branch :+1:

serima:
Thank you!

PR #167, opened 2018-04-23
use PHPUnit 7 as well

DQNEO:
https://phpunit.de/

for PHP7.1 and 7.2, it's good to allow them to use PHPUnit7.


markstory:
Thanks!

PR #168, opened 2018-05-16
Clear out the microseconds when setting time.

markstory:
Refs #106

### DIFF
--- a/src/Traits/FrozenTimeTrait.php
+++ b/src/Traits/FrozenTimeTrait.php
@@ -61,7 +61,7 @@ trait FrozenTimeTrait
      */
     public function setTime($hours, $minutes, $seconds = null, $microseconds = null)
     {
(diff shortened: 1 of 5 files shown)


src/Traits/FrozenTimeTrait.php
```diff
@@ -61,6 +61,10 @@ protected function stripTime($time)
      */
     public function setTime($hours, $minutes, $seconds = null, $microseconds = null)
     {
+        if (version_compare(PHP_VERSION, '7.1.0', '>=')) {
```
lorenzo: can we make this a constant somewhere? Would like to avoid the comparison for each object created

markstory: We could define a class/global constant inside of `src/carbon_compat.php`?

lorenzo: Sounds good

lorenzo:
Looks like we need to suppress a warning

PR #169, opened 2018-05-19
Make 'test now' a singular thing.

markstory:
Having to manage and align 4 test now instances was pretty awkward and led to annoying issues like cakephp/cakephp#11908. By having only one test now we avoid that issue entirely as there is only one place for 'now' to live.

Refs cakephp/cakephp#11908


dereuromark:
Looks like modify() should have been part of the interface.

Issue #165, opened 2018-04-05
Handle DateInterval Roll-Over


votemike:
It would be useful if there was a way to handle DateInterval Roll-Over.
For example, I would hope that `(ChronosInterval::seconds(107))->format('%I:%S')` would output `01:47` rather than `00:107`.

othercorey:
Is there anything left to do on this?

votemike:
I have just tested version 1.2.8 and the rollover still happens.

ADmad:
Guys this is an issue, not a pull request. So nothing has actually changed :slightly_smiling_face:.

ADmad:
PR #172 does mention it *could* fix this issue but I guess that's not the case.

othercorey:
@ADmad I'm aware it's an issue.  That's why I was asking why it was still open if #172 handles the roll-over.  

@votemike You are saying it is still **working** correct?

votemike:
No. #172 does not fix the problem.
My request remains open/valid/unfixed.

othercorey:
Thanks for the clarification. I misunderstood the merged pull request then.

ajibarra:
While having this at seconds, minutes, hours and days level could make sense, when weeks and months are involved it is impossible to actually know the amount of weeks or months without knowing the actual start date. 

It will lead to bad calculations if you enter 35 days. According to this it should create an interval with 1 months and..5 days? but then if you add this interval to 05/15 the result will be 06/20 but if the interval would have been created as 35 days the result would be 06/19. That's the reason DateInterval base class works that way.

So again, it makes sense for seconds minutes and hours and it is something we could improve but I feel it does not make sense the effort while this is something that can be easily done at app level. 

To summarize I think this issue can be closed.

PR #172, opened 2018-06-17
[Improvement] - ChronosInterval - Display interval in readable format

tatitati:
# Why this improvement
So far ChronosInterval displays any duration that you want but in a very basic way. So durations like ```P7000D3000S``` (7.000 days and 3.000 seconds) are totally fine with the current implementation, but makes little sense for the normal person.

So with this improvement instead of having: ```P1835D``` (1.835 days) we will have ```P5Y10D``` (5 years and 10 days), for example.

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -503,8 +503,7 @@ class ChronosInterval extends DateInterval
             static::PERIOD_MINUTES => $i,
             static::PERIOD_SECONDS => $s,
         ]);
-
-
+        
         $specString = static::PERIOD_PREFIX;
 
         foreach ($date as $key => $value) {
@@ -523,6 +522,5 @@ class ChronosInterval extends DateInterval
         }
 
         return $this->invert === 1 ? '-' . $specString : $specString;
-
     }
 }
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -436,17 +436,74 @@ class ChronosInterval extends DateInterval
      */
     public function __toString()
     {
+        // equivalence
+        $oneMinuteInSeconds = 60;
+        $oneHourInSeconds = $oneMinuteInSeconds * 60;
+        $oneDayInSeconds = $oneHourInSeconds * 24;
+        $oneMonthInDdays = 365/12;
+        $oneMonthInSeconds = $oneDayInSeconds * $oneMonthInDdays;
+        $oneYearInSeconds = 12 * $oneMonthInSeconds;
+
+        // convert
+        $ySecs = $this->y * $oneYearInSeconds;
+        $mSecs = $this->m * $oneMonthInSeconds;
+        $dSecs = $this->d * $oneDayInSeconds;
+        $hSecs = $this->h * $oneHourInSeconds;
+        $iSecs = $this->i * $oneMinuteInSeconds;
+        $sSecs = $this->s;
+
+        $totalSecs = $ySecs + $mSecs + $dSecs + $hSecs + $iSecs + $sSecs;
+
+        $y = null;
+        $m = null;
+        $d = null;
+        $h = null;
+        $i = null;
+
+        // years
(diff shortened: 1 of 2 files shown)


dereuromark:
Impressive refactor work. Makes sense :)

src/ChronosInterval.php
```diff
@@ -436,16 +436,72 @@ public function add(DateInterval $interval)
      */
     public function __toString()
     {
+        // equivalence
+        $oneMinuteInSeconds = 60;
+        $oneHourInSeconds = $oneMinuteInSeconds * 60;
+        $oneDayInSeconds = $oneHourInSeconds * 24;
+        $oneMonthInDdays = 365 / 12;
```
markstory: Looks like an extra `D` in days.

tests/Interval/IntervalToStringTest.php
```diff
@@ -25,51 +25,102 @@ public function testZeroInterval()
...
+        $this->assertEquals('P1D', (string)$ci3);
+        $this->assertEquals('P1D', (string)$ci4);
     }
 
     public function testMixedDateInterval()
     {
-        $ci = new ChronosInterval(1, 2, 0, 3);
-        $this->assertEquals('P1Y2M3D', (string)$ci);
+        $ci1 = new ChronosInterval(1, 2, 0, 3);
+        $ci2 = new ChronosInterval(0, 14, 0, 3);
```
votemike: Do you want to have a test where a smaller time measurement bleeds in to the next one? Such as `new ChronosInterval(1, 2, 0, 3, 36);`?

votemike:
Do we _only_ want to do this sort of logic in the `__toString()` method? I think this sort of logic would also be useful if I were to pass 60 seconds in to the constructor and be able to call `$this->i` and get 1 minute. I'm tempted to do this logic in the constructor. Or at least have a method that can be called to get this "saner" version of the object.

votemike:
The other thing to remember is that this will change the outcome of method. Which may cause issues for people expecting the old outcome. So it will need to be a breaking change (if the change to `__toString()` is kept like this).

votemike:
Also, this looks like it would fix https://github.com/cakephp/chronos/issues/165

markstory:
@votemike What makes you think this is a breaking change? No return type or method signature has changed. Changing the data representation should be acceptable in a bugfix/minor release.

> Do we only want to do this sort of logic in the __toString() method? I think this sort of logic would also be useful if I were to pass 60 seconds in to the constructor

I think for now only changing `__toString()` is reasonable. Changing the constructor would also require changing all the setter methods so that the entire object is representative of standard wall clocks, which increased the risk of change.

votemike:
@markstory as the returned string will change. People may be depending on `(string)(new ChronosInterval(0, 12));` returning "P12M" rather than "P1Y".

Instead of changing the `__toString()` method. What about a new method that returns a "processed" object? That would avoid any changes to the output of existing methods. And it would allow all existing methods to work take advantage of these changes.

Something like:
```
public function getSaneObject()
{
    // figure out correct y, m, d, h, i, s
    return new static($years, $months, $weeks, $days, $hours, $minutes, $seconds);
}
```

And then you'd be able to do:
```
    $interval = new ChronosInterval(0, 12);
    echo $interval; // "P12M" (output unchanged from before PR)
    echo $interval->y; // "0" (output unchanged from before PR)
    echo $interval->getSaneObject(); // "P1Y" (desired output due to PR)
    echo $interval->getSaneObject()->y; // "1" (desired output due to PR)
```

markstory:
Is your concern that folks will be doing assertions or operations on the return of the string cast? Given that the forms are functionally equivalent when fed back into a DateInterval, I don't think I understand the practical risk.

Adding methods like `getSaneObject()` seems less than idea. Perhaps a `normalize()` method that normalizes the internal data?

votemike:
Yes. That was my concern. I don't know how big of an impact it would have.

`normalize()` is a much better name for the method.

PR #179, opened 2018-10-05
Update shorthand functions to point to the full named functions

richardhughes:
Updated the shorthand functions to point to the fully named functions so its clear when reading code what each function does. 

All tests pass with this change 

```
PHPUnit 7.4.0 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 12:
  - Element 'phpunit', attribute 'syntaxCheck': The attribute 'syntaxCheck' is not allowed.

  Test results may not be as expected.


.............................................................   61 / 1285 (  4%)
.............................................................  122 / 1285 (  9%)
.............................................................  183 / 1285 ( 14%)
.............................................................  244 / 1285 ( 18%)
.............................................................  305 / 1285 ( 23%)
.............................................................  366 / 1285 ( 28%)
.............................................................  427 / 1285 ( 33%)
.............................................................  488 / 1285 ( 37%)
.............................................................  549 / 1285 ( 42%)
.............................................................  610 / 1285 ( 47%)
.............................................................  671 / 1285 ( 52%)
.............................................................  732 / 1285 ( 56%)
.............................................................  793 / 1285 ( 61%)
.............................................................  854 / 1285 ( 66%)
.............................................................  915 / 1285 ( 71%)
.............................................................  976 / 1285 ( 75%)
............................................................. 1037 / 1285 ( 80%)
............................................................. 1098 / 1285 ( 85%)
............................................................. 1159 / 1285 ( 90%)
............................................................. 1220 / 1285 ( 94%)
............................................................. 1281 / 1285 ( 99%)
....                                                          1285 / 1285 (100%)

Time: 519 ms, Memory: 18.00MB

OK (1285 tests, 3667 assertions)

```

### DIFF
--- a/tests/DateTime/ComparisonTest.php
+++ b/tests/DateTime/ComparisonTest.php
@@ -41,7 +41,7 @@ class ComparisonTest extends TestCase
      */
     public function testEqualToTrue($class)
     {
-        $this->assertTrue($class::createFromDate(2000, 1, 1)->eq($class::createFromDate(2000, 1, 1)));
+        $this->assertTrue($class::createFromDate(2000, 1, 1)->equalTo($class::createFromDate(2000, 1, 1)));
     }
 
     /**
@@ -50,7 +50,7 @@ class ComparisonTest extends TestCase
      */
     public function testEqualToFalse($class)
     {
-        $this->assertFalse($class::createFromDate(2000, 1, 1)->eq($class::createFromDate(2000, 1, 2)));
+        $this->assertFalse($class::createFromDate(2000, 1, 1)->equalTo($class::createFromDate(2000, 1, 2)));
     }
 
     /**
@@ -59,7 +59,7 @@ class ComparisonTest extends TestCase
      */
     public function testEqualWithTimezoneTrue($class)
     {
-        $this->assertTrue($class::create(2000, 1, 1, 12, 0, 0, 'America/Toronto')->eq($class::create(
+        $this->assertTrue($class::create(2000, 1, 1, 12, 0, 0, 'America/Toronto')->equalTo($class::create(
             2000,
             1,
             1,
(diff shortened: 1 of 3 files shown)


dereuromark:
Looks good, just needs a cs fix:
https://travis-ci.org/cakephp/chronos/jobs/437829718

richardhughes:
Cheers, should be fixed now! :+1:

lorenzo:
I would prefer duplication of the implementation, at least for some time, if we plan to remove the other functions. One of the features of chromosomes is that it avoids some indirection to optimize for performance.

tests/DateTime/ComparisonTest.php
```diff
@@ -41,7 +41,7 @@ public function testGetSetWeekendDays($class)
      */
     public function testEqualToTrue($class)
     {
-        $this->assertTrue($class::createFromDate(2000, 1, 1)->eq($class::createFromDate(2000, 1, 1)));
+        $this->assertTrue($class::createFromDate(2000, 1, 1)->equalTo($class::createFromDate(2000, 1, 1)));
```
markstory: Shouldn't we keep coverage on the old methods as well?

ravage84: If we plan to duplicate the functionality, certainly. If not, as currently proposed, it wouldn,t hurt.

ravage84:
> One of the features of chromosomes is that it avoids some indirection to optimize for performance.

@lorenzo what do you mean by "chromosomes"?

lorenzo:
Chromosomes = chronos according to my iPhone

dereuromark:
Do we still want to merge this? We plan on making a new 1.3 minor release soon.

othercorey:
I think changing all the internal uses is odd.

Instead, I would expect to see a set of test cases that confirm the short hand == full name.  You can't just rely on implementation details to test.

```assertTrue($time->gt($otherTime) === $time->greaterThan($otherTime))```

othercorey:
Cleaned up PR.  Kept original comparison functions internally.  Added unit tests for full name comparisons.

othercorey:
I think the `equalTo()` helpers could be just `equals()` or `greaterThanOrEquals()`.  It lines up with other function names in cakephp.

PR #181, opened 2018-10-18
Don't set class aliases if carbon exists.

### LINKED ISSUE #180: This library breaks nesbot/carbon version 2.0+
This library does not seem to be compatible with Carbon 2.0 or higher due to the aliases defined in `src/carbon_compat.php`.  Attempting to use Carbon while this library is installed results in the following fatal error on PHP 7.2:

> PHP Fatal error:  Declaration of Carbon\Carbon::diffInDaysFiltered(Closure $callback, $date = NULL, $absolute = true) must be compatible with Cake\Chronos\ChronosInterface::diffInDaysFiltered(callable $callback, ?Cake\Chronos\ChronosInterface $dt = NULL, $abs = true) in /tmp/chronos/vendor/nesbot/carbon/src/Carbon/Carbon.php on line 489

And this error on PHP 7.1:

> PHP Fatal error:  Declaration of Carbon\Carbon::eq($date): bool must be compatible with Cake\Chronos\ChronosInterface::eq(Cake\Chronos\ChronosInterface $dt) in /tmp/chronos/vendor/nesbot/carbon/src/Carbon/Carbon.php on line 489

I can replicate this by performing the following steps:

1. Clone this repository
2. Run `composer require --dev nesbot/carbon` (installing any version 2.0.0 or higher)
3. Run the following PHP code:

```php
require_once 'vendor/autoload.php';

$d = \Carbon\Carbon::now();
```

I would not expect the presence of one library (Chronos) to completely break another (Carbon).

Commenting-out the last two lines in `src/carbon_compat.php` resolves this issue for me.  Perhaps these lines should only be conditionally executed if Carbon is not present?

markstory:
Fixes #180

### DIFF
--- a/src/carbon_compat.php
+++ b/src/carbon_compat.php
@@ -12,7 +12,9 @@
  */
 define('CHRONOS_SUPPORTS_MICROSECONDS', version_compare(PHP_VERSION, '7.1.0', '>='));
 
-// Create class aliases for Carbon so applications
-// can upgrade more easily.
-class_alias('Cake\Chronos\Chronos', 'Carbon\MutableDateTime');
-class_alias('Cake\Chronos\ChronosInterface', 'Carbon\CarbonInterface');
+if (!class_exists('Carbon\Carbon')) {
+    // Create class aliases for Carbon so applications
+    // can upgrade more easily.
+    class_alias('Cake\Chronos\Chronos', 'Carbon\MutableDateTime');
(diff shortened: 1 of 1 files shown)


colinodell:
Thank you for the quick fix!  Looking forward to getting this into my project.

lorenzo:
Looks good!

PR #184, opened 2019-01-25
Add DifferenceFormatterInterface.

ADmad:
`Cake\I18n\RelativeTimeFormatter` can now implement this interface.


dereuromark:
If we dont add php71 typehints we could already introduce it in 1.x series.

ADmad:
It can be backported to 1.x after this is merged.

PR #186, opened 2019-01-27
2.x type hints

ADmad:


src/Chronos.php
```diff
@@ -134,7 +134,7 @@ public function toMutable(): MutableDateTime
      *
      * @return $this
      */
-    public function copy()
+    public function copy(): ChronosInterface
```
dereuromark: Same thing as in core: We should not add the typehints for chainable, it breaks usability.
Instead use the Spryker sniffs to ensure this is all clean and safe, and that solves it properly.

dereuromark:
Rest looks good.
But we sure overuse the traits here a bit IMO :)

src/Chronos.php
ADmad: Like I said on core PR, show me how it breaks usability.

That said the docblock of `ChronosInterface::copy()` has `@return static` and since chronos is supposed to be immutable shouldn't this method return a `clone $this` instead of `$this`?

markstory:
>  But we sure overuse the traits here a bit IMO :)

The alternative was 5K line classes with duplication between the immutable and mutable variants.

ADmad:
Ping

src/Chronos.php
ADmad: I modified `Chronos::copy()` to return cloned instance.

Issue #164, opened 2018-03-19
Error: debugInfo must return an array


hktr92:
Hello, there.

While debugging with phpstorm and xdebug the behavior of a controller which uploads avatar images, I came across an error thrown by PHP:
![image](https://user-images.githubusercontent.com/7689151/37600954-7c9960b0-2b91-11e8-8391-489cdc185347.png)

I think that it's related to #129 which introduced the given method. 
It's a Symfony 4-based project and for dev environment, I'm using PHP 7.1.15 with Xdebug 2.6.0. Tried an `composer update cakephp/chronos`, but still fails.

What can be done to fix the encountered issue?

markstory:
But it does return an array https://github.com/cakephp/chronos/blob/be20ca1679c532556c9abec67f230273bc874ef4/src/Chronos.php#L136-L144

hktr92:
Created a new directory, ran `composer req cakephp\chronos`, instantiated the class, stepped into and xdebug died. Removed the timezone line from array, stepped into and it worked, but with each step over, i got this error(s) in console:

    /usr/bin/php7.1 -dxdebug.remote_enable=1 -dxdebug.remote_mode=req -dxdebug.remote_port=9000 -dxdebug.remote_host=127.0.0.1 /home/hktr92/Playground/xdebug-crash/exec.php                                       
    PHP Warning:  DateTimeImmutable::getTimezone(): The DateTime object has not been correctly initialized by its constructor in /home/hktr92/Playground/xdebug-crash/vendor/cakephp/chronos/src/Chronos.php on lin
    PHP Stack trace:                                                                                                                                                                                               
    PHP   1. {main}() /home/hktr92/Playground/xdebug-crash/exec.php:0                                                                                                                                              
    PHP   2. Cake\Chronos\Chronos->__construct($time = *uninitialized*, $tz = *uninitialized*) /home/hktr92/Playground/xdebug-crash/exec.php:8                                                                     
    PHP   3. Cake\Chronos\Chronos->__debugInfo() /home/hktr92/Playground/xdebug-crash/vendor/cakephp/chronos/src/Chronos.php:80                                                                                    
    PHP   4. Cake\Chronos\Chronos->getTimezone() /home/hktr92/Playground/xdebug-crash/vendor/cakephp/chronos/src/Chronos.php:169                                                                                   
[...]

PR #188, opened 2019-02-10
Fix Error: debugInfo must return an array #164

odan:
### Description

When you try to inspect a variable with Xdebug, the magic method `__debugInfo()` is getting called because Xdebug tries to load the variable via `var_dump()`. If the internal state of your object is not initialized (depends on the breakpoint), you receive errors like this:

* `Warning: DateTimeImmutable::format(): The DateTime object has not been correctly initialized by its constructor in vendor\cakephp\chronos\src\Chronos.php on line 195`
* `Warning: DateTimeImmutable::getTimezone(): The DateTime object has not been correctly initialized by its constructor`

The reason is, that a unfinished (Chronos) object has no `date` and `timezone` property set and the current `__debugInfo()` works only for fully a initialized Chronos objects.

### The fix

The fix checks whether the object properties exist. If not, the affected functions are not executed.

### Tests

It's technically impossible (for me) to cover this PR by unit tests, because it's not possible to simulate xdebug breakpoints with phpunit. I have tested this PR with PhpStorm and Xdebug to make sure the fix works.

### General thoughts on the use of `__debugInfo`

Using `__debugInfo` is a problem for all XDebug users because it causes debugging and stability issues. In the long run, the only sensible solution would be to remove the magic method `__debugInfo` from the complete Chronos codebase. If you think so, I would be happy to create a PR.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -44,8 +44,8 @@ use DateTimeZone;
  * @property-read bool $dst daylight savings time indicator, true if DST, false otherwise
  * @property-read bool $local checks if the timezone is local, true if local, false otherwise
  * @property-read bool $utc checks if the timezone is UTC, true if UTC, false otherwise
- * @property-read string  $timezoneName
- * @property-read string  $tzName
+ * @property-read string $timezoneName
+ * @property-read string $tzName
  */
 class Chronos extends DateTimeImmutable implements ChronosInterface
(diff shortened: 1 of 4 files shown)


dereuromark:
Looks good now.

markstory:
Thank you ❤️

PR #193, opened 2019-05-04
Upgrade to PHP 7.2 and PHPUnit 8.0.

ADmad:


markstory:
We need to take 7.1 out of the build matrix to get a passing build.

markstory:
Merging as the phpcs build can be updated separately.

PR #195, opened 2019-05-10
Fix opis closure serialization issues

steinkel:
While working with opis/closure package I found some issues serializing MutableDateTime objects, see the reference ticket https://github.com/opis/closure/issues/31 and @msarca pointed me to this direction for fixing the setter at MutableDateTime and allow opis to serialize and unserialize the object correctly.

Here would be another possibly related reference https://bugs.php.net/bug.php?id=75232.


steinkel:
hmm, should we keep BC with php 5.5 as of today?

dereuromark:
5.6 should suffice in next minor IMO.

src/MutableDateTime.php
```diff
@@ -47,6 +47,8 @@
  * @property-read bool $utc checks if the timezone is UTC, true if UTC, false otherwise
  * @property-read string $timezoneName
  * @property-read string $tzName
+ * @property-read string $date used in opis/closure serialization via reflection
+ * @property-read int $timezone_type used in opis/closure serialization via reflection
```
dereuromark: Those are only for test though?

steinkel: yes, and used internally via Reflection by opis/closure

dereuromark: But only in require-dev (testing). IMO that must not leak into the actual classes, only confuses to have those there and unused, right?
You could make a TestMutableDateTime that adds them on top, for example.

steinkel: https://github.com/opis/closure/blob/master/src/SerializableClosure.php#L403 there

steinkel: OK, I'll create test classes and keep Stan happy, any concerns with the actual solution before fixing Stan and php5.5 (hopefully) ?

markstory: I am confused. How do these new properties fix serialization? Are we adding setters for properties that were readonly before?

steinkel: @markstory the serialization done by opis/closure relies in Reflection to extract the instance properties, https://gist.github.com/steinkel/80bfdb2755cbc8cc722c1ab7559f398a and Reflection is returning "date", "timezone_type", as ReflectionProperty, possibly because of > https://bugs.php.net/bug.php?id=75232. We don't want or need these new setter exposed in docs (will get this fixed as suggested by @dereuromark), but we would need them to allow opis work with this class (or any other class extending \DateTime)

dereuromark:
Looks good to me now.

steinkel:
OK the remaining issue is related to php5.5, we we're going to remove it in the next minor I don't think it's worth to fix...

dereuromark:
We have to first fix https://github.com/cakephp/chronos/issues/201, then we could merge and release this.

othercorey:
With php 7.4, those properties are no longer exposed.

othercorey:
These test cases don't make sense for Chronos.  They would need to be split up into the parts that expect to be set.

PR #198, opened 2019-05-30
Test now times

dakota:
When using the test now, and creating a time object with a time, the test date is ignored.

This changes that behaviour.

### DIFF
--- a/src/Traits/RelativeKeywordTrait.php
+++ b/src/Traits/RelativeKeywordTrait.php
@@ -30,7 +30,7 @@ trait RelativeKeywordTrait
     public static function hasRelativeKeywords($time)
     {
         // Just a time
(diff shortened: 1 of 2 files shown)


src/Traits/RelativeKeywordTrait.php
```diff
@@ -29,6 +29,11 @@ trait RelativeKeywordTrait
      */
     public static function hasRelativeKeywords($time)
     {
+        // Just a time
+        if (preg_match('/^[0-2][0-9]:[0-5][0-9](?::[0-5][0-9])?$/', $time) > 0) {
```
dereuromark: usually `!== false` here - or just if (!

dakota: Good point, change pushed.

PR #204, opened 2019-06-16
Fix relative parsing with timezones.

markstory:
When parsing relative times + timezones the intent is to have a relative time in the provided timezone. This means we don't want to apply a timezone adjustment before creating the final result.

Fixes #203

Issue #203, opened 2019-06-16
Further BC issue after #202


joshbmarshall:
There's an issue with timezones when parsing a time by itself. Introduced by #198 and partially fixed.
For example I add to the TestingAidsTest.php file the following test, it fails:

	public function testParseRelativeWithTimezoneAndTestValueSet($class) {
		$notNow = $class::parse('2013-07-01 12:00:00', 'America/New_York');
		$class::setTestNow($notNow);

		$this->assertSame('06:30:00', $class::parse('2013-07-01 06:30:00', 'America/Mexico_City')->toTimeString());
		$this->assertSame('06:30:00', $class::parse('6:30', 'America/Mexico_City')->toTimeString());

		$this->assertSame('2013-07-01T06:30:00-04:00', $class::parse('2013-07-01 06:30:00')->toIso8601String());
		$this->assertSame('2013-07-01T06:30:00-04:00', $class::parse('6:30')->toIso8601String());
		$this->assertSame('2013-07-01T06:30:00-05:00', $class::parse('2013-07-01 06:30:00', 'America/Mexico_City')->toIso8601String());
		$this->assertSame('2013-07-01T06:30:00-05:00', $class::parse('6:30', 'America/Mexico_City')->toIso8601String());
		$this->assertSame('2013-07-01T06:30:00-05:00', $class::parse('6:30:00', 'America/Mexico_City')->toIso8601String());
		$this->assertSame('2013-07-01T06:30:00-05:00', $class::parse('06:30:00', 'America/Mexico_City')->toIso8601String());
	}

Instead it is returning an offset time based on the timezone. In the above example it is returning 05:30:00 instead of 06:30:00 because of the 1 hour timezone difference.

markstory:
I'm not sure this problem is related to the changes in #198 as the provided tests fail on a different line even without the changes in #198. I do agree that the the provided tests should pass though.

markstory:
Pull request up now.

joshbmarshall:
Hi @markstory I have commented on your pull request it causes another issue.
The problem with time parsing has come from #198 and has caused other issues with further tests along the track. It's likely the time detection should be in a separate function. Now I see how you have done it perhaps I can create a pull request?

PR #205, opened 2019-06-17
Fix relative and fixed time parsing with timezones

joshbmarshall:
Complete fix for #203 - includes pull request #204 by @markstory 
Separated the test for a time string to a relative time keywords (e.g. 5 minutes ago) and only avoid applying timezone variances for fixed time-of-day strings.

### DIFF
--- a/src/Traits/RelativeKeywordTrait.php
+++ b/src/Traits/RelativeKeywordTrait.php
@@ -26,7 +26,7 @@ trait RelativeKeywordTrait
      * @param string $time The time string to check.
      * @return bool true if there is a keyword, otherwise false
      */
-    public static function hasRelativeTime($time)
+    private static function isTimeExpression($time)
     {
         // Just a time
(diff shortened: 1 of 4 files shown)


src/Traits/RelativeKeywordTrait.php
```diff
@@ -21,19 +21,33 @@ trait RelativeKeywordTrait
...
     /**
-     * Determine if there is a relative keyword in the time string, this is to
-     * create dates relative to now for test instances. e.g.: next tuesday
+     * Determine if there is just a time in the time string
      *
      * @param string $time The time string to check.
      * @return bool true if there is a keyword, otherwise false
      */
-    public static function hasRelativeKeywords($time)
+    public static function hasRelativeTime($time)
```
markstory: Does this new method need to be public? This method checks that the input is a time, would its name better match what it does if it were called `isTimeExpression()` or `isOnlyTime()`?

joshbmarshall: Good point. I've renamed to `isTimeExpression()` and made private. I had made it public as that is what `hasRelativeKeywords()` was.

PR #206, opened 2019-07-26
Improve documentation

bancer:


src/Traits/FactoryTrait.php
```diff
@@ -50,7 +50,7 @@ public static function instance(DateTimeInterface $dt)
      * (new Chronos('Monday next week'))->fn()
      *
      * @param string $time The strtotime compatible string to parse
-     * @param \DateTimeZone|string|null $tz The DateTimeZone object or timezone name.
+     * @param \DateTimeInterface|\DateTimeZone|string|null| $tz The DateTimeZone object or timezone name.
```
garas: Last `|` looks wrong

src/Traits/FactoryTrait.php
```diff
@@ -50,7 +50,7 @@ public static function instance(DateTimeInterface $dt)
      * (new Chronos('Monday next week'))->fn()
      *
      * @param string $time The strtotime compatible string to parse
-     * @param \DateTimeZone|string|null $tz The DateTimeZone object or timezone name.
+     * @param \DateTimeInterface|\DateTimeZone|string|null $tz The DateTimeZone object or timezone name.
```
ADmad: You probably meant to change type of first argument, not second.

dereuromark:
We should try to get phpstan to level 5, this would help find some of those issues.

bancer:
Scrutinizer also detects them.

dereuromark:
We should stop 5.5 in travis matrix IMO. 5.6 at this point should be minimum.

PR #210, opened 2019-10-20
Use Chronos instead of Date in hour/minute components example

othercorey:
Using Date is confusing as setting those components has no effect.


markstory:
Thank you 👍

PR #215, opened 2019-10-30
2.x - Added 7.4snapshot php config

othercorey:
Adding 7.4snapshot php config to travis matrix.  This seems to be missing from a previous merge.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -193,21 +193,11 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
      */
     public function __debugInfo(): array
     {
-        // Conditionally add properties if state exists to avoid
-        // errors when using a debugger.
-        $vars = get_object_vars($this);
-
         $properties = [
             'hasFixedNow' => static::hasTestNow(),
+            'time' => $this->format('Y-m-d H:i:s.u'),
+            'timezone' => $this->getTimezone()->getName(),
         ];
-
-        if (isset($vars['date'])) {
-            $properties['time'] = $this->format('Y-m-d H:i:s.u');
-        }
-
-        if (isset($vars['timezone'])) {
-            $properties['timezone'] = $this->getTimezone()->getName();
-        }
 
         return $properties;
     }
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: php
 php:
   - 7.2
   - 7.3
+  - 7.4snapshot
 
 sudo: false
 
@@ -21,10 +22,13 @@ matrix:
     - php: 7.2
       env: COVERAGE=1 DEFAULT=0
 
+  allow_failures:
+    - php: 7.4snapshot
+
   fast_finish: true
 
 before_install:
-  - phpenv config-rm xdebug.ini
+  - if [[ $TRAVIS_PHP_VERSION != "7.4snapshot" ]]; then phpenv config-rm xdebug.ini; fi
 
 install:
   - travis_retry composer install --no-interaction --prefer-dist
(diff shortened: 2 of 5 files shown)


dereuromark:
7.4 is already in master, but in allow_failures
maybe we need a merge into 2.x instead.

dereuromark:
Would be nice if we didnt need the allow failure part anymore for master and 2.x
Especially since 7.4 is now becoming stable in a few weeks.

othercorey:
Probably makes sense to investigate that on master.  Does cakephp 3.x plan to support php 7.4 or leave it as failing?

ADmad:
Though here the failures on PHP seem relevant and not related to phpunit, sobthey should be addressed.

othercorey:
This is what I get for fixing the merge lol.  I'll poke around at it.

ADmad:
No good deed goes unpunished :P

othercorey:
Resolved.  The array returned by get_object_vars() on DateTime objects changed.  This was probably unreliable at best.  I build the properties directly instead of conditionally.  This is the same way cakephp builds the debug properties for I18n Time based on Chronos.

src/Chronos.php
```diff
@@ -193,22 +193,12 @@ public static function hasTestNow(): bool
      */
     public function __debugInfo(): array
     {
-        // Conditionally add properties if state exists to avoid
-        // errors when using a debugger.
-        $vars = get_object_vars($this);
```
markstory: How will the new code work when paused in a debugger? The conditional property building was to work around debuggers needing to format Chronos objects before the constructor was finished. See #188

othercorey: How does I18n\Time work?

https://github.com/cakephp/cakephp/blob/d509a3a820582087c4b3afa8670d00dd600dd191/src/I18n/DateFormatTrait.php#L448

othercorey: Let me give a more useful comment :).

**7.3**
```
php > print_r(get_object_vars(new \DateTime('now')));
Array
(
    [date] => 2019-10-31 03:10:01.434797
    [timezone_type] => 3
    [timezone] => UTC
)
```

**7.4**
```
php > print_r(get_object_vars(new \DateTime('now')));
Array
(
)
```

othercorey: We could get really tricky and have a property that tracks whether the parent constructor was called. This would need to exist in Chronos and CakePHP.

garas: @markstory Since Xdebug 2.7.2 won't use `__debugInfo()` for user-defined classes, Xdebug will show real properties.

dakota: That is a real pity :( Should we open a issue on xdebug that `__debugInfo()` no longer works?

garas: `serialize($this)` equals to `O:20:"Cake\Chronos\Chronos":0:{}` if not object initiated.

Check if it ends with `:0:{}`.

saeideng: @dakota 
https://github.com/xdebug/xdebug/pull/469

othercorey: Maybe the question we should be asking is does Date really need __debuginfo().

markstory: Well if xdebug is no longer calling `__debugInfo()` then we won't have the old problem and these changes will be 👌

othercorey:
Do we expect 1.x to work with PHP 7.4?  Should I back-port the fix and remove 7.4 from travis failing matrix?

src/Chronos.php
othercorey: Looks like Xdebug 2.8 is the first release that supports php 7.4.

markstory:
@othercorey Yes. Thank you for taking care of the backport :clap:

PR #216, opened 2019-10-30
Removed nightly from php build matrix

othercorey:
Nightly tries to run PHP 8.0.0-dev now which isn't supported by anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ php:
   - 7.1
   - 7.3
   - 7.4snapshot
-  - nightly
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


othercorey:
I wouldn't exactly call nightly "running", but it does actually try to run.  This at least syncs up with 2.x branch.

garas:
`Installing phpunit/phpunit (4.1.6): Downloading (100%)` 🤣

`composer install` could be executed with `--ignore-platform-reqs` probably on `nightly`.

ADmad:
Just get rid of `nightly` form the matrix, it was only required to test on 7.4 during it's early days.

othercorey:
@ADmad I was hoping someone would say that.  It tries to run PHP 8.0.0-dev now.  I was just following on a previous commit that fixes a bad 2.x merge.

Issue #149, opened 2017-07-17
Different date is returned for Chronos::today($tz) and Date::today($tz)


stil:
Currently here is 18 Jul, 1:31 AM in Europe/Warsaw (UTC+2:00).
For some reason ```Chronos::today()``` and ```Date::today()``` return different date.

```php
Chronos::today("Europe/Warsaw")
```
```
Cake\Chronos\Chronos {#306
  +"time": "2017-07-18 00:00:00.000000"
  +"timezone": "Europe/Warsaw"
  +"hasFixedNow": false
}
```

```php
Date::today("Europe/Warsaw")
```
```
Cake\Chronos\Date {#306
  +"date": "2017-07-17"
  +"hasFixedNow": false
}
```

Am I missing something or is it a bug?

markstory:
Date objects always use UTC as their timezone, ignoring the input timezone. In this scenario it might make sense to get the date in Warsaw and then create another instance in utc for the same calendar day.

davispuh:
> ```
> Date::today("Europe/Warsaw")
> ```

The problem with `Date` is that it doesn't have any particular timezone, but it should have it. There can be a day difference due to timezones. For example comparing `Date` from `PDT` and `Date` from `AEST`.
So that above is same as `Date::today(new DateTimeZone('UTC'))` which obviously won't match `Chronos($tz)` because `Date` always ignores `tz`.

Currently using `Date` with different timezones is real PITA.

```php
$timestamp = 1565679600; // 2019-08-13 00:00:00 PDT

// need to do this to get date of specific timezone from unix timestamp
$timezone1 = new DateTimeZone('America/Los_Angeles');
$dateTime1 = Chronos::createFromTimestamp($timestamp, $timezone1);
$date1 = Date::instance($dateTime1);

$timezone2 = new DateTimeZone('Australia/Sydney');
$date2 = Date::instance(Chronos::createFromTimestamp($timestamp, $timezone2));

// we want to compare if 2 dates from different timezones are on same day
if ($date1 == $date2) {
  // they're on same day
} else {
  // different days
}

// we want to compare Date to DateTime where each has different timezone
// but we can't use $date2 directly because it doesn't have info about it's timezone
$date2WithTimezone = Chronos::createFromTimestamp($date2->getTimestamp(), $timezone2);
if ($date2WithTimezone > $dateTime1) {

}

```

markstory:
[...]

PR #218, opened 2019-10-30
2.x - Allow timezone parameter for Date/MutableDate construction

othercorey:
https://github.com/cakephp/chronos/issues/149

This allows specifying the timezone for a time passed to Date and MutableDate constructors.  The timezone is used to grab the date before storing it in UTC as before. 

With this change Chronos::today(timezone) and Date::today(timezone) will match (assuming time() hasn't adjusted past a date change of course).

This also fixes a couple missing test case issues where the time was not set to 00:00:00 for Date instances such as using a fixed date + relative time: '2001-01-01 +10 minutes'

The timezone for TestNow instances is also adjusted before the relative time change to support things like 'midnight' being correct in the target timezone (e.g. 00:00:00 regardless of timezone).  This matches PHP's DateTime behavior when passing such values with a timezone parameter.

### DIFF
--- a/src/Traits/RelativeKeywordTrait.php
+++ b/src/Traits/RelativeKeywordTrait.php
@@ -36,7 +36,7 @@ trait RelativeKeywordTrait
     private static function isTimeExpression($time)
     {
         // Just a time
-        if (is_string($time) && preg_match('/^[0-2]?[0-9]:[0-5][0-9](?::[0-5][0-9])?$/', $time)) {
+        if (is_string($time) && preg_match('/^[0-2]?[0-9]:[0-5][0-9](?::[0-5][0-9](?:\.[0-9]{1,6})?)?$/', $time)) {
             return true;
         }
 
@@ -61,5 +61,21 @@ trait RelativeKeywordTrait
         }
 
         return false;
+    }
+
+    /**
+     * Determines if there is no fixed date in the time string.
+     *
+     * @param \DateTimeInterface|string|null $time The time string to check
+     * @return bool true if doesn't contain a fixed date
+     */
+    private static function isRelativeOnly($time): bool
+    {
+        if (!is_string($time)) {
+            return false;
+        }
+
+        // must not contain fixed date before relative keywords or time expression
+        return preg_match('/[0-9]{4}-[0-9]{1,2}-[0-9]{1,2}/', $time) !== 1;
     }
 }
--- a/src/Traits/RelativeKeywordTrait.php
+++ b/src/Traits/RelativeKeywordTrait.php
@@ -71,6 +71,10 @@ trait RelativeKeywordTrait
      */
     private static function isRelativeOnly($time): bool
     {
+        if ($time === null) {
+            return true;
+        }
+
         if (!is_string($time)) {
             return false;
         }
--- a/tests/Date/ConstructTest.php
+++ b/tests/Date/ConstructTest.php
@@ -303,12 +303,7 @@ class ConstructTest extends TestCase
         $this->assertEquals($london->format('Y-m-d'), $c->format('Y-m-d'));
         $this->assertSame('UTC', $c->tzName);
 
-        // Confirm that timezone is adjusted before midnight is set.
-        // This implies that the relative time used by Date helpers
-        // is affected by the timezone parameter properly.
-        $london = new DateTimeImmutable('midnight', $londonTimezone);
-        $this->assertEquals('00:00:00', $london->format('H:i:s'));
-
+        // now adjusted to London time
         $c = $class::today($londonTimezone);
         $this->assertEquals(Chronos::today($londonTimezone)->format('Y-m-d'), $c->format('Y-m-d'));
 
--- a/tests/Date/ConstructTest.php
+++ b/tests/Date/ConstructTest.php
@@ -13,9 +13,12 @@ declare(strict_types=1);
  */
 namespace Cake\Chronos\Test\Date;
 
+use Cake\Chronos\Chronos;
 use Cake\Chronos\Date;
 use Cake\Chronos\MutableDate;
 use Cake\Chronos\Test\TestCase;
+use DateTimeImmutable;
+use DateTimeZone;
 
 /**
  * Test constructors for Date objects.
@@ -27,18 +30,6 @@ class ConstructTest extends TestCase
      * @return void
      */
     public function testCreateFromTimestamp($class)
-    {
-        $ts = 1454284800;
-        $date = $class::createFromTimestamp($ts);
-        $this->assertEquals('UTC', $date->tzName);
-        $this->assertEquals('2016-02-01', $date->format('Y-m-d'));
-    }
-
-    /**
-     * @dataProvider dateClassProvider
-     * @return void
-     */
-    public function testCreateFromTimestampUtc($class)
     {
         $ts = 1454284800;
         $date = $class::createFromTimestamp($ts);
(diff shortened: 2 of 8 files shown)


othercorey:
I am working on testing cakephp 4.x against this branch.  There is some default/constructor change missing I believe.

tests/Date/ConstructTest.php
```diff
@@ -258,5 +249,110 @@ public function testConstructWithRelative($class)
...
+        $c = new $class('now', $londonTimezone);
+        $london = new DateTimeImmutable('now', $londonTimezone);
+        $this->assertEquals($london->format('Y-m-d'), $c->format('Y-m-d'));
+        $this->assertSame('UTC', $c->tzName);
+
+        // Confirm that timezone is adjusted before midnight is set.
+        // This implies that the relative time used by Date helpers
+        // is affected by the timezone parameter properly.
+        $london = new DateTimeImmutable('midnight', $londonTimezone);
+        $this->assertEquals('00:00:00', $london->format('H:i:s'));
```
markstory: When would this fail? It looks like it is testing PHP.

tests/Date/ConstructTest.php
```diff
@@ -258,5 +249,110 @@ public function testConstructWithRelative($class)
...
+        $this->assertEquals('2010-01-04 00:00:00', $c->format('Y-m-d H:i:s'));
+        $this->assertSame('UTC', $c->tzName);
+
+        // TestNow is adjusted to London time
+        $c = $class::today($londonTimezone);
+        $this->assertEquals('2010-01-02 00:00:00', $c->format('Y-m-d H:i:s'));
+        $this->assertEquals(Chronos::today($londonTimezone)->format('Y-m-d'), $c->format('Y-m-d'));
+        $this->assertSame('UTC', $c->tzName);
+
+        // TestNow is ajusted to London time
```
markstory:
```suggestion
        // TestNow is adjusted to London time
```

tests/Date/ConstructTest.php
othercorey: This is me being paranoid about testing.  This *is* testing PHP, but it's testing the premise behind Date using the timezone parameter.

I wanted to find a way to test Date with 'now' or 'today' without using TestNow.

othercorey: Oh, when would it fail?  It would fail if PHP was doing: now > midnight > timezone.  That would mean local time is adjusted to 00:00:00 and then adjusted to the target timezone. 

This confirms it is doing: now > timezone > midnight.

othercorey:
@markstory Sorry for all the pushes. I was doing a lot of testing.  

I'm getting a strange error in an unrelated test at the moment.  It seems DateTime::diff() is returning 4 weeks for the last month.

```
1) Cake\Chronos\Test\DateTime\DiffTest::testDiffForHumansRelative with data set #2 (Cake\Chronos\Chronos Object (...), Cake\Chronos\Chronos Object (...), '11 months before')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'11 months before'
+'10 months before'
```

Was able to complete a test ahead cakephp 4.x branch.  I was missing support for default/null.

There is an "invalid" test in cakephp where I18n\Date assumes UTC time and expects +23hrs is the same day regardless of local time.  This is relying on an implementation detail of Chronos\Date which is now properly working off local time for relative values.

https://github.com/cakephp/cakephp/blob/f505b7a41cfe7dcb30e09cee7bc62e0645974bfd/tests/TestCase/I18n/DateTest.php#L421

othercorey:
The weird 1 month - 4 weeks issue resolved itself. I think it's an edge case from testing literally at the end of the month (2019-10-31 23:00:00 - 24:00:00).

I'm not sure how to force the tests to re-run, but they pass on my fork.

https://travis-ci.org/othercorey/chronos/builds/605775613

tests/Date/ConstructTest.php
markstory: I'm not sure we need to be paranoid about how PHP is working, unless we're using it as a frame of reference to compare our results.

othercorey: I can remove it.  It was placeholder while I was testing everything else.

othercorey:
@markstory I ran the new benchmark suite against this change.  So far, looks like there are no significant cost differences.

**Before**

```
+-------------------------+---------+------+-----+------------+----------+----------+----------+----------+
| subject                 | set     | revs | its | mem_peak   | best     | mean     | mode     | worst    |
+-------------------------+---------+------+-----+------------+----------+----------+----------+----------+
| benchNow                | date    | 1000 | 5   | 2,483,800b | 5.920μs  | 6.262μs  | 6.025μs  | 7.227μs  |
| benchNowTimezone        | date    | 1000 | 5   | 2,483,816b | 5.734μs  | 6.058μs  | 5.874μs  | 6.802μs  |
| benchRelative           | date    | 1000 | 5   | 2,483,800b | 7.753μs  | 7.995μs  | 8.076μs  | 8.167μs  |
| benchRelativeTimezone   | date    | 1000 | 5   | 2,483,816b | 7.952μs  | 8.096μs  | 8.043μs  | 8.340μs  |
| benchFixed              | date    | 1000 | 5   | 2,483,800b | 5.641μs  | 5.736μs  | 5.769μs  | 5.841μs  |
| benchFixedTimezone      | date    | 1000 | 5   | 2,483,816b | 5.645μs  | 5.800μs  | 5.725μs  | 6.057μs  |
```

**After**

```
+-------------------------+---------+------+-----+------------+----------+----------+----------+----------+
| subject                 | set     | revs | its | mem_peak   | best     | mean     | mode     | worst    |
+-------------------------+---------+------+-----+------------+----------+----------+----------+----------+
| benchNow                | date    | 1000 | 5   | 2,484,776b | 5.751μs  | 6.221μs  | 5.934μs  | 7.391μs  |
| benchNowTimezone        | date    | 1000 | 5   | 2,484,792b | 5.820μs  | 5.948μs  | 6.020μs  | 6.055μs  |
| benchRelative           | date    | 1000 | 5   | 2,484,776b | 7.913μs  | 8.004μs  | 8.030μs  | 8.076μs  |
| benchRelativeTimezone   | date    | 1000 | 5   | 2,484,792b | 7.635μs  | 8.044μs  | 7.929μs  | 8.567μs  |
| benchFixed              | date    | 1000 | 5   | 2,484,776b | 5.323μs  | 5.683μs  | 5.743μs  | 5.947μs  |
| benchFixedTimezone      | date    | 1000 | 5   | 2,484,792b | 5.250μs  | 5.681μs  | 5.591μs  | 6.241μs  |
```

src/Traits/RelativeKeywordTrait.php
```diff
@@ -62,4 +62,24 @@ public static function hasRelativeKeywords(?string $time): bool
...
+        if ($time === null) {
+            return true;
+        }
+
+        if (!is_string($time)) {
+            return false;
+        }
+
+        // must not contain fixed date before relative keywords or time expression
+        return preg_match('/[0-9]{4}-[0-9]{1,2}-[0-9]{1,2}/', $time) !== 1;
```
markstory: How does this pattern check for relative keywords?

othercorey: It does not.  This checks for an explicit date.  Anything other than an fixed date is assumed to be relative to TestNow.

Strings such as '2001-01-01 +10 days' will be sent to the constructor as they are not relative.

PR #219, opened 2019-11-01
Fixed error using get_object_vars on DateTime object in php 7.4

othercorey:
Backported https://github.com/cakephp/chronos/pull/215


dereuromark:
Nice work!

Issue #223, opened 2019-11-03
2.x - ChronosInterval doesn't support fractional/microseconds


othercorey:
PHP 7.1 added some support for fractional seconds in `DateInterval`.  

https://www.php.net/manual/en/class.dateinterval.php

They added a `$f` property which stores microseconds as a decimal fraction.  So if the interval is 2.5 seconds then `$s` is `2` and `$f` is `0.5`.  This is populated when an interval is created from diffing two DateTime instances.

However, DateInterval's constructor does not support fractional seconds in the spec string.

There are several ways to add support for getting microseconds:

- Simply expose $f via 'microseconds' property and leave it as a fractional second
- Expose $f via 'microseconds' but convert it into an integer number of microseconds.
- Add $f fractional seconds to $s when returning 'seconds'

Adding support for setting microseconds is probably limited:

- Allow setting $f via 'microseconds' as a fractional second
- Allow setting $f via 'microseconds' as an integer number of microseconds
- Allow setting $f via 'seconds' containing seconds and fractional second together

markstory:
Exposing the `f` property as a fractional using and `microseconds` as an integer seems reasonable to me. It lets us be compatible with PHP's DateInterval and also improve the interface with our extensions.

PR #224, opened 2019-11-06
Added microseconds support to ChronosInterval.

othercorey:
https://github.com/cakephp/chronos/issues/223

This does not add microseconds to `__toString()`.  There is an expectation that the spec string created can be passed to DateInterval constructor.  Unfortunately, DateInterval does not support decimals in the spec string contrary to the ISO spec.

We could add it as decimal seconds and simply not include microseconds in the unit test.  I'm not sure that's a good situation though.

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -50,7 +50,7 @@ use InvalidArgumentException;
  * @method static \Cake\Chronos\ChronosInterval seconds($seconds = 1) Create instance specifying a number of seconds.
  * @method static \Cake\Chronos\ChronosInterval second($seconds = 1) Alias for seconds
  * @method static \Cake\Chronos\ChronosInterval microseconds($microseconds = 1) Create instance specifying a number of microseconds.
- * @method static \Cake\Chronos\ChronosInterval microsecond($microseconds = 1) Alias for seconds
+ * @method static \Cake\Chronos\ChronosInterval microsecond($microseconds = 1) Alias for microseconds
  */
 class ChronosInterval extends DateInterval
 {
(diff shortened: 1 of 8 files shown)


src/ChronosInterval.php
```diff
@@ -48,6 +49,8 @@
  * @method static \Cake\Chronos\ChronosInterval minute($minutes = 1) Alias for minutes
  * @method static \Cake\Chronos\ChronosInterval seconds($seconds = 1) Create instance specifying a number of seconds.
  * @method static \Cake\Chronos\ChronosInterval second($seconds = 1) Alias for seconds
+ * @method static \Cake\Chronos\ChronosInterval microseconds($microseconds = 1) Create instance specifying a number of microseconds.
+ * @method static \Cake\Chronos\ChronosInterval microsecond($microseconds = 1) Alias for seconds
```
markstory:
```suggestion
 * @method static \Cake\Chronos\ChronosInterval microsecond($microseconds = 1) Alias for microseconds
```

src/ChronosInterval.php
```diff
@@ -256,22 +266,16 @@ public function __get(string $name)
...
-            case 'days':
-                return $this->days;
-            case 'y':
-            case 'm':
-            case 'd':
-            case 'h':
-            case 'i':
-            case 's':
-            case 'invert':
-                return parent::__get($name);
```
markstory: Why is this being removed?

othercorey: These public properties exist on DateInterval.  This isn't required for the PR to work, but it confused me when investigating where to add ->f and ->microseconds support.

Issue #220, opened 2019-11-02
FactoryTrait::create() and createFromTime() missing microseconds


othercorey:
These interfaces seem to be missing microsecond parameters.

I assume there will be backwards compatibility issues, but this seems like a strange omission.  

Since $tz has no type, it could accept the microsecond value, but it will hit a runtime failure instead of silently converting/ignoring.

The type hints in 2.x should ensue any internal code is using it right.

> @param \DateTimeZone|string|null $tz The DateTimeZone object or timezone name the new instance should use.

```
    public static function create(
        ?int $year = null,
        ?int $month = null,
        ?int $day = null,
        ?int $hour = null,
        ?int $minute = null,
        ?int $second = null,
        $tz = null
    ): ChronosInterface {
```

markstory:
> Since $tz has no type, it could accept the microsecond value, but it will hit a runtime failure instead of silently converting/ignoring.

You could count the number of arguments and argument types to determine whether or not you have a timezone or a microsecond. The timezone should be a string or a `DateTimeZone` while microseconds should be an integer.

dereuromark:
PR open

PR #226, opened 2019-11-12
2.x - Added microseconds to FactoryTrait::create()

othercorey:
I also changed the format letters used by create() to the standard Y-m-d to match every other place in cakephp. They are aliases of each other.

Fixes https://github.com/cakephp/chronos/issues/220

### DIFF
--- a/src/Traits/FactoryTrait.php
+++ b/src/Traits/FactoryTrait.php
@@ -134,9 +134,10 @@ trait FactoryTrait
      * will be used.
      *
      * If $hour is null it will be set to its now() value and the default values
-     * for $minute, $second and $microsecond will be their now() values.
-     * If $hour is not null then the default values for $minute and $second
+     * for $minute and $second will be their now() values. The default for milliseconds
      * will be 0.
+     * If $hour is not null then the default values for $minute, $second
+     * and $microsecond will be 0.
      *
      * @param int|null $year The year to create an instance with.
      * @param int|null $month The month to create an instance with.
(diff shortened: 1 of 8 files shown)


src/Traits/FactoryTrait.php
```diff
@@ -154,24 +155,27 @@ public static function create(
         ?int $hour = null,
         ?int $minute = null,
         ?int $second = null,
+        ?int $microsecond = null,
```
othercorey: For users that already pass a timezone, this will throw a type error. $tz will never be an integer and if $microsecond isn't an integer then the type check fails.

tests/DateTime/CreateFromTimeTest.php
```diff
@@ -71,14 +71,25 @@ public function testCreateFromTimeWithSecond($class)
         $this->assertSame(21, $d->second);
     }
 
+    /**
+     * @dataProvider classNameProvider
+     * @return void
+     */
+    public function testCreateFromTimeWithMicroecond($class)
```
garas:
```suggestion
    public function testCreateFromTimeWithMicrosecond($class)
```

src/Traits/FactoryTrait.php
```diff
@@ -134,7 +134,7 @@ public static function minValue(): ChronosInterface
      * will be used.
      *
      * If $hour is null it will be set to its now() value and the default values
-     * for $minute and $second will be their now() values.
+     * for $minute, $second and $microsecond will be their now() values.
```
garas: This is lie because `date('u')` will always generate 000000.

src/Traits/FactoryTrait.php
```diff
@@ -154,24 +155,27 @@ public static function create(
...
-        $day = $day ?? date('j');
+        $month = $month ?? date('m');
+        $day = $day ?? date('d');
 
         if ($hour === null) {
-            $hour = date('G');
+            $hour = date('H');
             $minute = $minute ?? date('i');
             $second = $second ?? date('s');
+            $microsecond = $microsecond ?? date('u');
```
garas: All these parts should use same `$now` timestamp.

It could be `2:59:59.999999` while taking hour, but then `3:00:00.000000` while taking minute/second/millisecond, and final time might be 2:00:00.000000`.

garas:
Since PHP 7.1, `DateTime::__construct` microseconds are filled with actual value. Not with '00000'. https://www.php.net/manual/en/datetime.construct.php

This could be rewritten using `DateTime`, as `date()` doesn't support milliseconds.

othercorey:
> This could be rewritten using `DateTime`, as `date()` doesn't support milliseconds.

I wasn't sure if it was necessary, but I can change it.

garas:
Initial commit tried to get milliseconds from current time, and it should get it.

`$now` would be `DateTime` instance, all `date()` would become `$now->format()` and it should be done.

othercorey:
> Initial commit tried to get milliseconds from current time, and it should get it, and the lie would become truth.
> 
> `$now` would be `DateTime` instance, all `date()` would become `$now->format()` and it should be done.

Sounds good. I had to update/remove a few unit tests that required the second to not change which is invalid.  This caused failures when adding microseconds which always change.

PR #227, opened 2019-11-14
Date should use the system timezone

markstory:
While pinning calendar dates to UTC is helpful in many situations it makes Date objects hard to use when the server local timezone is far away from UTC (like Asia/Tokyo) as Date::today() and other factory methods will create the wrong date for half the day. By using the server's default timezone we can have more correct date generation in these timezones. The behavior will be unchanged for applications that have their default timezone set to UTC.

### DIFF
--- a/src/Date.php
+++ b/src/Date.php
@@ -82,7 +82,7 @@ class Date extends DateTimeImmutable implements ChronosInterface
      *
      * Date instances lack time components, however due to limitations in PHP's
      * internal Datetime object the time will always be set to 00:00:00, and the
-     * timezone will always be UTC. Normalizing the timezone allows for
+     * timezone will always be the server local time. Normalizing the timezone allows for
      * subtraction/addition to have deterministic results.
      *
      * @param \DateTime|\DateTimeImmutable|string|int|null $time Fixed or relative time
@@ -97,7 +97,7 @@ class Date extends DateTimeImmutable implements ChronosInterface
         $testNow = Chronos::getTestNow();
         if ($testNow === null || !static::isRelativeOnly($time)) {
             $time = $this->stripTime($time, $tz);
-            parent::__construct($time, new DateTimeZone('UTC'));
+            parent::__construct($time, new DateTimeZone(date_default_timezone_get()));
 
             return;
         }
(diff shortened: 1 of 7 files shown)


src/Date.php
```diff
@@ -82,7 +82,7 @@ class Date extends DateTimeImmutable implements ChronosInterface
      *
      * Date instances lack time components, however due to limitations in PHP's
      * internal Datetime object the time will always be set to 00:00:00, and the
-     * timezone will always be UTC. Normalizing the timezone allows for
+     * timezone will always be the server local time. Normalizing the timezone allows for
```
othercorey: Need to copy this to MutableDate

tests/Date/ConstructTest.php
```diff
@@ -63,9 +63,13 @@ public function testCreateFromEmptyWithTestNow($class)
      */
     public function testCreateFromTimestamp($class)
     {
+        $tz = date_default_timezone_get();
+        date_default_timezone_set('Europe/London');
```
othercorey: Why change this one specifically?

othercorey:
Can you copy the isToday unit test for Date + MutableDate?

src/Date.php
```diff
@@ -97,7 +97,7 @@ public function __construct($time = 'now', $tz = null)
         $testNow = Chronos::getTestNow();
         if ($testNow === null || !static::isRelativeOnly($time)) {
             $time = $this->stripTime($time, $tz);
-            parent::__construct($time, new DateTimeZone('UTC'));
+            parent::__construct($time, new DateTimeZone(date_default_timezone_get()));
```
othercorey: Remove `new DateTimeZone` to optimize default time zone in DateTime constructor.

tests/Date/ConstructTest.php
markstory: I wanted a non UTC timezone that wasn't on a different day from the test's existing assertion. This could be any European timezone, I just chose london arbitrarily.

ADmad: Without DST `Europe/London` is same as UTC isn't it?

othercorey: When using a timestamp, the date will not be modified regardless of local time zone.

src/Date.php
markstory: Can't do that, I get errors from PHP.  `TypeError: DateTime::__construct() expects parameter 2 to be DateTimeZone, string given` and `TypeError: DateTimeImmutable::__construct() expects parameter 2 to be DateTimeZone, string given`

othercorey: You don't pass a 2nd parameter. It will use the default automatically.

markstory:
> Can you copy the isToday unit test for Date + MutableDate?

I think I added the tests you were suggesting, could you take a look?

Issue #176, opened 2018-07-23
Accept an Array from datetime Inputs for DateTime::__construct() in MutableDateTime.php


3811015:
I leave this for a discussion. 
I think it would be a small enhancement, which is easy to achive and worth it...

Imagine, you are building your own validation rule in cakephp... Let's keep it simple, something like a datatime-field is not permitted to be more than one hour in the future.
You will end up building something like this:

```
$validator->add('begin', 'costum', [
   'rule' => function ($value, $context){
        //Calculate the diff against Time::now()
        //if diff is ok, return true
        return false;
   },
   'message' => __('Der Beginn liegt mehr als eine Stunde in der Zukunft.')
]);
...
return $validator;
```

If you look at $value, you will see an array like:

```
[
	'year' => '2017',
	'month' => '03',
	'day' => '02',
	'hour' => '15',
	'minute' => '30'
]
```

So you can't do following:
```
$valueTime = new Time($value);
$valueDiff = $value->diff(Time::now());
if()...
```

You will need to write your stupid own array to string conversion in this validator. You will need to write your own tests for this array to string conversion. Would be much nicer, to put the array in there and not worrying about all else so much.

Kind regards.

saeideng:
cake uses 
https://github.com/cakephp/cakephp/blob/1f51a72d4b85ab98e0a1dcc195d50565c847724e/src/Validation/Validation.php#L1568-L1591

PR #234, opened 2019-11-26
2.x - Added createFromArray()

othercorey:
This is a copy of the array to string conversion in `DateTimeType::marshal()` and `Validation::_getDateString()` from cakephp.  We can just construct the instance using this helper in `marshal()` and maintain the format here.

Can back-port to 1.3 without the microseconds support.

Closes https://github.com/cakephp/chronos/issues/176

### DIFF
--- a/src/Traits/FactoryTrait.php
+++ b/src/Traits/FactoryTrait.php
@@ -252,6 +252,11 @@ trait FactoryTrait
     /**
      * Creates a ChronosInterface instance from an array of date and time values.
      *
+     * The 'year', 'month' and 'day' values must all be set for a date. The time
+     * values all default to 0.
+     *
+     * The 'timezone' value can be any format supported by `\DateTimeZone`.
+     *
      * Allowed values:
      *  - year
      *  - month
@@ -260,7 +265,7 @@ trait FactoryTrait
      *  - minute
      *  - second
      *  - microsecond
-     *  - meridian ('am or 'pm')
+     *  - meridian ('am' or 'pm')
      *  - timezone
      *
      * @param (int|string)[] $values Array of date and time values.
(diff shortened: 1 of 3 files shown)


src/Traits/FactoryTrait.php
```diff
@@ -249,6 +249,58 @@ public static function createFromFormat($format, $time, $tz = null): ChronosInte
...
+        $formatted = '';
+        if (
+            isset($values['year'], $values['month'], $values['day']) &&
+            (
+                is_numeric($values['year']) &&
+                is_numeric($values['month']) &&
+                is_numeric($values['day'])
+            )
+        ) {
+            $formatted .= sprintf('%d-%02d-%02d', $values['year'], $values['month'], $values['day']);
```
markstory: Should year be forced to 4 digits to support pre 1000 years?

othercorey: Definitely.

src/Traits/FactoryTrait.php
```diff
@@ -249,6 +249,63 @@ public static function createFromFormat($format, $time, $tz = null): ChronosInte
...
+
+        if (isset($values['meridian']) && (int)$values['hour'] === 12) {
+            $values['hour'] = 0;
+        }
+        if (isset($values['meridian'])) {
+            $values['hour'] = strtolower($values['meridian']) === 'am' ? $values['hour'] : $values['hour'] + 12;
+        }
+        $formatted .= sprintf(
+            '%s%02d:%02d:%02d.%06d',
+            empty($formatted) ? '' : ' ',
```
garas: This space exists when date set, move this space up to above `sprintf` of date.

othercorey: I think the original code did this to avoid extra spaces when there is only date.  I can change it if you want.

garas: I know it came from original code. It would be one less condition and lower code complexity as there is either _date time_ or _time_, so space better inserted after _date_ part, not conditionally before _time_ part.

othercorey: You're right.  Fixed!

PR #240, opened 2020-01-08
Add dayOfWeekName support

Wouter0100:
This lets us use the magic property dayOfWeekName to get the weekday name using `date('l');`.

I will be adding some tests asap.


dereuromark:
https://github.com/cakephp/chronos/pull/250

PR #243, opened 2020-01-29
Fix modify method signature in ChronosInterface

B-Galati:
static is interpreted as a static method in phpdoc, not as a return type.

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -19,7 +19,7 @@ use DateTimeInterface;
 /**
  * An extension to the DateTimeInterface for a friendlier API
  *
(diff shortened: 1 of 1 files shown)


src/ChronosInterface.php
```diff
@@ -19,7 +19,7 @@
 /**
  * An extension to the DateTimeInterface for a friendlier API
  *
- * @method static modify(string $relative)
+ * @method ChronosInterface modify(string $relative)
```
markstory:
```suggestion
 * @method \Cake\Chronos\ChronosInterface modify(string $relative)
```

PR #244, opened 2020-01-30
Fixed CS errors. Updated phpstan to 0.12

othercorey:


othercorey:
Our unit tests fail due to a 1 month vs 4 weeks time difference at the end of the month.

PR #245, opened 2020-01-31
Allow creating Chronos instances using `new` from existing Chronos instances

pableu:
Convert `ChronosInterface` instances to string in `Chronos::_construct`. 

Passing a Chronos instance to `Chronos::parse` or `new Chronos()` was possible in 1.x (where it was typecast to string silently), but not longer in 2.x because we now have `strict_types=1`.

I know I could use `Chronos::instance`, but we have some existing usages in our code where we pass either a string or an existing Chronos instance to `Chronos::parse` (or `new Chronos()`).

This worked in 1.3: 

    use \Cake\Chronos\Chronos;
    new Chronos(new Chronos());

But in 2.x it says `DateTimeImmutable::__construct() expects parameter 1 to be string, object given`.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -96,8 +96,8 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
         static::$_lastErrors = [];
         $testNow = static::getTestNow();
         if ($testNow === null) {
-            if ($time instanceof ChronosInterface) {
-                $time = $time->__toString();
+            if ($time instanceof \DateTimeInterface) {
+                $time = $time->format(ChronosInterface::DEFAULT_TO_STRING_FORMAT);
             }
             parent::__construct($time ?? 'now', $tz);
 
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -97,7 +97,7 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
         $testNow = static::getTestNow();
         if ($testNow === null) {
             if ($time instanceof ChronosInterface) {
-              $time = $time->__toString();
+                $time = $time->__toString();
             }
             parent::__construct($time ?? 'now', $tz);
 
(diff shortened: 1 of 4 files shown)


src/Chronos.php
```diff
@@ -96,6 +96,9 @@ public function __construct($time = 'now', $tz = null)
         static::$_lastErrors = [];
         $testNow = static::getTestNow();
         if ($testNow === null) {
+            if ($time instanceof ChronosInterface) {
+                $time = $time->__toString();
```
markstory:
```suggestion
                $time = (string)$time;
```

othercorey:
This should allow any `DateTimeInterface` the same as `Chronos\Date`.

pableu:
Thanks for your feedback. I added conversion from any `DateTimeInterface` to `\Cake\Chronos\Chronos` and `\Cake\Chronos\MutableDateTime`, and also some tests. What do you think?

markstory:
Thanks!

src/Chronos.php
```diff
@@ -96,6 +96,9 @@ public function __construct($time = 'now', $tz = null)
         static::$_lastErrors = [];
         $testNow = static::getTestNow();
         if ($testNow === null) {
+            if ($time instanceof \DateTimeInterface) {
+                $time = $time->format(ChronosInterface::DEFAULT_TO_STRING_FORMAT);
```
othercorey: This doesn't include microseconds.

tests/Date/ConstructTest.php
```diff
@@ -377,4 +377,35 @@ public function testConstructWithLargeTimezoneChange($class)
...
+    /**
+     * @dataProvider dateClassProvider
+     * @return void
+     */
+    public function testCreateFromDateTimeInterface($class)
+    {
+        $existingClass = new \DateTimeImmutable();
+        $newClass = new $class($existingClass);
+        self::assertInstanceOf($class, $newClass);
+        self::assertEquals($existingClass->format('Y-m-d 00:00:00'), $newClass->format('Y-m-d H:i:s'));
```
othercorey: Also missing microseconds.

pableu:
@othercorey you are absolutely right, I totally missed that. I can try to add it. Should I do that in a new PR or can we reopen this one?

othercorey:
It should be separate. This is merged.

PR #248, opened 2020-02-13
Respect microseconds when constructing from DateTimeInterface

pableu:
As rightly pointed out by @othercorey in #245, we should also consider the microseconds when constructing Chronos or MutableDateTime from an existing DateTimeInterface.

This PR fixes that.


markstory:
Thanks!

PR #264, opened 2020-05-30
Added missing @var tags

othercorey:


dereuromark:
For constants I dont really see the need for the tags.
As here, by design the type cannot really dynamically change. Most constants are literally set into stone.
The few that could be extended would only ever change inside the same type (e.g. limit 1000 to 100), never in type.


But since they also don't really hurt, fair enough :)

PR #267, opened 2020-08-13
Consistency for assertion methods call

peter279k:
# Changed log

- The PHPUnit accepts two approaches about calling assertions. And there're two approaches used on this repository.
- It seems that using the `$this` to call assertion times is greater than using `self::` to call assertions.

To be consistency, it should let all PHPUnit assertion methods call to be the `$this`.


dereuromark:
In general I would also like to see assertSame() instead of assertEquals() for scalar values.

peter279k:
> In general I would also like to see assertSame() instead of assertEquals() for scalar values.

Thanks. And it will be another PR :).

PR #269, opened 2020-08-14
Replace assertEquals with assertSame

peter279k:
# Changed log

- It's related to this [comment](https://github.com/cakephp/chronos/pull/267#issuecomment-673383878).
- Using the `assertSame` assertion to replace `assertEquals` and it makes all value equal checking strictly.
  - When using the `assertSame`, it cannot assert the different objects are same strictly.
  - To resolve above this issue, it should cast object to string and assert their strings are same.

### DIFF
--- a/src/Traits/FormattingTrait.php
+++ b/src/Traits/FormattingTrait.php
@@ -233,7 +233,7 @@ trait FormattingTrait
      */
     public function toQuarter(bool $range = false)
     {
-        $quarter = ceil($this->format('m') / 3);
+        $quarter = (int)ceil($this->format('m') / 3);
         if ($range === false) {
             return $quarter;
         }
--- a/tests/TestCase/DateTime/StringsTest.php
+++ b/tests/TestCase/DateTime/StringsTest.php
@@ -236,10 +236,10 @@ class StringsTest extends TestCase
     public function testToUnixString($class)
     {
         $time = $class::parse('2014-04-20 08:00:00');
-        $this->assertEquals('1397995200', $time->toUnixString());
+        $this->assertSame('1397995200', $time->toUnixString());
 
         $time = $class::parse('2021-12-11 07:00:01');
-        $this->assertEquals('1639224001', $time->toUnixString());
+        $this->assertSame('1639224001', $time->toUnixString());
     }
 
     /**
@@ -268,7 +268,7 @@ class StringsTest extends TestCase
      */
     public function testToQuarter($date, $expected, $range = false)
     {
-        $this->assertEquals($expected, (new Chronos($date))->toQuarter($range));
+        $this->assertSame($expected, (new Chronos($date))->toQuarter($range));
     }
 
     /**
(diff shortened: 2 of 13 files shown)


dereuromark:
I was only referring to non objects 
The object identity checks with assertEquals dont check for same objects though, but values of them.
So I think those are good too.

tests/TestCase/DateTime/StringsTest.php
```diff
@@ -250,9 +250,9 @@ public function testToUnixString($class)
     public function toQuarterProvider()
     {
         return [
-            ['2007-12-25', 4],
-            ['2007-9-25', 3],
-            ['2007-3-25', 1],
+            ['2007-12-25', 4.0],
+            ['2007-9-25', 3.0],
+            ['2007-3-25', 1.0],
```
dereuromark, line 255: I think this reveals a defect now. See documented return type of method.

peter279k: Could you make suggested change about documented return type of method?

The `toQuarterProvider` data provider seems to be the mixed type of `$expected`.

dereuromark: I meant the method you are testing here :)

peter279k: Ok. Thanks and sorry for my misunderstanding.

If having any request change, please feel free to let me know :).

peter279k: I change the returned type hints for `FormattingTrait::toQuarter()` method.

Please see this latest diff on this PR :).

src/Traits/FormattingTrait.php
```diff
@@ -229,7 +229,7 @@ public function toUnixString(): string
      * Returns the quarter
      *
      * @param bool $range Range.
-     * @return int|array 1, 2, 3, or 4 quarter of year or array if $range true
+     * @return float|array 1, 2, 3, or 4 quarter of year or array if $range true
      */
     public function toQuarter(bool $range = false)
     {
```
peter279k, line 235: Just notice that the `ceil` function will always return `float` type, and it should use `float|array` type hints instead.

dereuromark: What I was going to say is that "int" is expected and "float" is wrongly returned :)
I think this is a bug because both from expectation and code using this, the float value makes no sense whatsoever.

peter279k: Thanks for your reply, @dereuromark.

To resolve this issue, I think we should cast the `ceil` function to be `int` and revert returned type hints to be `int|array`.

What do you think about this :)?

peter279k: @dereuromark, the above approach I have been updated on the latest PR commit, please feel free to review :).

dereuromark:
By the way: https://github.com/rectorphp/rector/pull/3958 :)

PR #270, opened 2020-08-14
Run testsuite on PHP 8.

ADmad:


ADmad:
I had to get rid of `phpbench/phpbench` as keeping it results in 
https://travis-ci.com/github/cakephp/chronos/jobs/372209955#L294

othercorey:
This was going to fail with the added test.

PR #276, opened 2020-08-28
Fix stripping of time from (Mutable)Date instance on PHP 8.

ADmad:


ADmad:
Hmm.. This change is not needed. Merging `master` into `2.next` fixed the failing test.

PR #290, opened 2021-04-02
Fix int/timestamp support in Chronos constructor

othercorey:
Chronos and MutableDateTime did not support the `int` typehint. This required cakephp Time and FrozenTime to wrap all timestamps first.

Changed the constructor typehints to `DateTimeInterface` as that's what users expect. This fixes a typehint mismatch between Chronos and wrapper libraries like cakephp.

@markstory @ADmad

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -84,11 +84,17 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
      * Please see the testing aids section (specifically static::setTestNow())
      * for more on the possibility of this constructor returning a test instance.
      *
-     * @param \DateTime|\DateTimeImmutable|string|int|null $time Fixed or relative time
+     * @param \DateTimeInterface|string|int|null $time Fixed or relative time
      * @param \DateTimeZone|string|null $tz The timezone for the instance
      */
     public function __construct($time = 'now', $tz = null)
     {
+        if (is_int($time)) {
+            parent::__construct('@' . $time);
+
+            return;
+        }
+
         if ($tz !== null) {
             $tz = $tz instanceof DateTimeZone ? $tz : new DateTimeZone($tz);
         }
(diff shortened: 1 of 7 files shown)


othercorey:
The failure is the test that randomly fails. I'm still looking into why.

src/Chronos.php
```diff
@@ -84,11 +84,17 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
      * Please see the testing aids section (specifically static::setTestNow())
      * for more on the possibility of this constructor returning a test instance.
      *
-     * @param \DateTime|\DateTimeImmutable|string|int|null $time Fixed or relative time
+     * @param \DateTimeInterface|string|int|null $time Fixed or relative time
      * @param \DateTimeZone|string|null $tz The timezone for the instance
      */
     public function __construct($time = 'now', $tz = null)
     {
+        if (is_int($time)) {
```
ADmad, line 92: It would be better to use `is_numeric()` to avoid BC break since that's what's used in core's datetime classes https://github.com/cakephp/cakephp/blob/0c658ec7aae862ce20e12f15d7c42d1ff58f9228/src/I18n/Time.php#L135 and PR https://github.com/cakephp/cakephp/pull/15421 intends to remove the constructors.

othercorey: This code needs some love. We have an implementation divergence now because of this.

Date/MutableDate use `is_int()` check in the code that eventually handles integer timestamps:
https://github.com/cakephp/chronos/blob/master/src/Traits/FrozenTimeTrait.php#L41

Of course, this code isn't obvious, but trying for a minimal fix first.

ADmad: Welp, in that case `is_int()` makes sense for consistency within the `chronos` package. We should retain the `is_numeric()` check in core classes to avoid BC break.

markstory (commented):
Looks good to me. We could remove the conditions in cake if we pin the required version higher.

PR #292, opened 2021-04-19
Improve Carbon Migration Guide in Readme

jojomi:
CarbonImmutable instances should be converted to Chronos instance (which are immutable), rather than the non-existant "ChronosImmutable".


markstory:
Thank you 👍

Issue #293, opened 2021-05-21
PHP 8.1: Cannot return ChronosInterface from DateTime methods


othercorey:
PHP 8.1 introduces more return types for internal methods including DateTime. Chronos currently returns `ChronosInterface` from some overriden DateTime methods. This is not compatible with the declared return type of `DateTime` or `DateTimeImmutable` because `ChronosInterface` does not extend those classes. 

```
Deprecated: Declaration of Cake\Chronos\Traits\FrozenTimeTrait::setTime($hours, $minutes, $seconds = null, $microseconds = null): Cake\Chronos\ChronosInterface should be compatible with DateTimeImmutable::setTime(int $hour, int $minute, int $second = 0, int $microsecond = 0): DateTimeImmutable in /home/runner/work/cakephp/cakephp/vendor/cakephp/chronos/src/Date.php on line 78
```

The only known solution is to return `static` from these methods instead of `ChronosInterface`. The `static` return type is available in PHP 8 only.

othercorey:
It's possible to suppress the deprecation with a `ReturnTypeWillChange` attribute.

othercorey:
Fixed in https://github.com/cakephp/chronos/pull/302

PR #297, opened 2021-05-25
Replace ChronosInterface return types

othercorey:
Refs https://github.com/cakephp/chronos/issues/293

* Replace `ChronosInterface` return type on non-static methods with `self`
* Replace missing return type on static methods with `static`
* Replace `mixed` return types on non-static methods with `self`

Previously, a lot of `@return` types were `static` instead of `self`. Only static methods need static binding type.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1']
+        php-version: ['8.0', '8.1']
         prefer-lowest: ['']
         include:
-          - php-version: '7.2'
+          - php-version: '8.0'
             prefer-lowest: 'prefer-lowest'
 
     steps:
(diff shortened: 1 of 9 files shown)


othercorey:
We have an issue where some functions accept `ChronosInterface` instance and return them with a `self` return type. This works ok at runtime because all of the instances are Chronos instances, but `ChronosInterface` does not fit the expected type.

.github/workflows/ci.yml
```diff
@@ -15,10 +15,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0', '8.1']
+        php-version: ['8.0', '8.1']
```
dereuromark, line 18: Why is PHP MIN now 8? Where did 7.2/7.3/7.4 go? At least the latter would be useful to keep IMO
Either way, this needs to be synced with composer.json min as well then.

othercorey: This is explained in the attached issue.

PR #298, opened 2021-05-25
Replace ChronosInterface return types with static

othercorey:
Refs https://github.com/cakephp/chronos/issues/293

This starts preparing Chronos for php 8.1 support which requires compatible return types for internal methods.

This replaces all return types with `static` except for `closest()`, `farthest()`, `min()` and `max()`. These functions return the `ChronosInterface` that was passed in which presents a static analysis issue. You can't show that `ChronosInterface` is `static`.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -147,7 +147,7 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
      *
      * @return static
      */
-    public function copy(): ChronosInterface
+    public function copy(): static
     {
         return clone $this;
     }
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -19,7 +19,7 @@ use DateTimeInterface;
 /**
  * An extension to the DateTimeInterface for a friendlier API
  *
- * @method \Cake\Chronos\ChronosInterface modify(string $relative)
+ * @method static modify(string $relative)
  */
 interface ChronosInterface extends DateTimeInterface
 {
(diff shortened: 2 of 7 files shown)


othercorey:
I'm not sure I understand the code sniff on return type. Why does a function returning an interface have to return an instance of self?

othercorey:
This seems to satisfy the 8.1 return types:

https://github.com/cakephp/cakephp/pull/15556/checks?check_run_id=2670329505

ADmad:
`static` type is available since PHP 7.4 so can you add it back to the build matrix.

othercorey:
> `static` type is available since PHP 7.4 so can you add it back to the build matrix.

I was working off this https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.static

ADmad:
I miss remembered, it's from 8.0 :(.

othercorey:
> I miss remembered, it's from 8.0 :(.

Just checked the RFC. It's definitely 8.0. This is pretty brutal. We'll have to maintain parity between 2.x and 3.x -- but thankfully that probably means 0 changes.

ADmad:
Yeah the silver lining is that the current version is pretty much baked, so rarely needs changes. Enhancements would go to a new branch.

ADmad:
Should we wait till 8.1 RC before merging this? It's early days for 8.1.

othercorey:
There is no rush. Please discuss how 4 functions that return the ChronosInterface they pass in should be handled.

My first thought was to fix the sniff.

PR #300, opened 2021-05-26
Add missing bool return type to comparison functions

othercorey:
I don't know why these were missing.

We might as well clean this sort of stuff up in a 3.0 release.


othercorey:
We can do this in 2.2 if we want since 3.0 is cancelled.

PR #317, opened 2022-07-26
Drop MutableDate and MutableDateTime classes

othercorey:
The docs need to be updated after this is approved.


ADmad (approved):

markstory:
I updated the main docs page and started off a 3.x migration guide.

PR #322, opened 2022-08-16
DifferenceTrait update to use getDiffFormatter()

andrii-pukhalevych:
Fix for cakephp/cakephp#16691

### DIFF
--- a/src/Traits/DifferenceTrait.php
+++ b/src/Traits/DifferenceTrait.php
@@ -308,25 +308,20 @@ trait DifferenceTrait
      */
     public function diffForHumans(?ChronosInterface $other = null, bool $absolute = false): string
     {
-        return static::diffFormatter()->diffForHumans($this, $other, $absolute);
+        return static::getDiffFormatter()->diffForHumans($this, $other, $absolute);
     }
 
     /**
-     * Get the difference formatter instance or overwrite the current one.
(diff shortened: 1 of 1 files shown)


src/Traits/DifferenceTrait.php
```diff
@@ -308,25 +308,20 @@ public static function fromNow($datetime)
...
 
     /**
-     * Get the difference formatter instance or overwrite the current one.
+     * Get the difference formatter instance.
      *
-     * @param \Cake\Chronos\DifferenceFormatterInterface|null $formatter The formatter instance when setting.
      * @return \Cake\Chronos\DifferenceFormatterInterface The formatter instance.
      */
-    public static function diffFormatter(?DifferenceFormatterInterface $formatter = null): DifferenceFormatterInterface
+    public static function getDiffFormatter(): DifferenceFormatterInterface
```
dereuromark, line 319: I dont think you can just replace a static API method like this.
You need to keep the old one for BC reasons.

andrii-pukhalevych: Regarding PHPUnit report, its left to add setDiffFormatter() method and update tests.
To preserve BC edit from maintainers will be needed :)

ADmad:
Chronos only generates English strings. The fix is needed in the core cakephp/cakephp#16692.

PR #323, opened 2022-09-05
Fix: Unexpected Human Interval Values

### LINKED ISSUE #307: diffForHumans() gives misleading results for future dates
Consider the following example of a future date two weeks ahead of today's date:

`echo Chronos::parse(new DateTime('2 weeks'))->diffForHumans();`

>1 week from now

It seems to me the expected out put should be _two weeks_ from now.

Also consider that as the current time ticks forward, the diff should probably not flip from two weeks to one week, since this is not very granular output. That is, once the difference becomes less than two weeks, it should probably change scale to days, e.g. 11 days, 10 days, etc. If it jumps straight from two weeks to one week, when should this occur? One second prior to the full two weeks? Half way between one week and two weeks? Up to one second after one full week? This presents a boundary condition problem that is difficult to reason about if we do not switch to a more granular time output, e.g. from weeks to days.

As an aside, the behaviour is not the same for past dates:

`echo Chronos::parse(new DateTime('-2 weeks'))->diffForHumans();`

>2 weeks ago

MGatner:
`diffForHumans()` is returning "not very human" intervals for unit values. Per the discussion this tries to identify a standard and  updates the values to match that.

Fixes #307


src/DifferenceFormatter.php
```diff
@@ -64,17 +64,17 @@ public function diffForHumans(
                 $unit = 'year';
                 $count = $diffInterval->y;
                 break;
-            case $diffInterval->m > 0:
+            case $diffInterval->days >= ChronosInterface::DAYS_PER_WEEK * 8:
```
markstory, line 67: In #307 you mentioned moving to months after 4 weeks not 8. What led to changing your mind?

MGatner: The table had 4 weeks but my recommendation was to give weeks a little bit more space:

> So here's my proposal: Increase days to be 1..21, give weeks a slight bump (like 3..6 or 3..8) and leave the rest alone.

I think at this point it's very subjective. If we want to stick to the same precision then we would go up to "60 weeks", which isn't very human :) If you prefer 4 weeks I'm glad to switch it; I just went with what sounds colloquial.

MGatner: @markstory checking back in here

markstory: Hmm, '60 weeks' is pretty long. To me (in Canada) 8 weeks is probably the right duration. I could see arguments for going to 3 months :shrug:

markstory:
Thanks @MGatner :clap:

Bilge:
Can we expect a tag soon?

othercorey:
This appears to be missing tests cases for the periods where the units should switch. For example at 8 weeks.

othercorey:
@markstory  After adding tests, this doesn't seem to work as expected. It goes from "7 weeks before" to "1 month before".

```php
[$now, $now->addWeeks(8), '2 months before'],
[$now, $now->addWeeks(7), '7 weeks before'],
```

```diff
--- Expected
+++ Actual
@@ @@
-'2 months before'
+'1 month before'
```

PR #324, opened 2022-09-05
Workflow PHP Version

MGatner:
Because static analysis tools don't run the code they are not tied to the actual PHP version and benefit from all the latest and greatest of newer PHP versions (in this case most keenly: performance).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
     - name: Setup PHP
       uses: shivammathur/setup-php@v2
       with:
-        php-version: '7.2'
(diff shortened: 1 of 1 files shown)


othercorey:
I think some of them default to detecting code is supported in the PHP version that's running. I think psalm now allows setting a profile. Don't know about phpstan.

In larger projects, we had issues creating a baseline that worked for multiple php versions, but doesn't seem to be an issue with just phpstan here.

MGatner:
Yes it should work fine; I've been doing this for years. If a problem arises you can specify a target version from the config file: https://phpstan.org/config-reference#phpversion

PR #326, opened 2022-09-27
Cleanup GA CI config

ADmad:


ADmad:
PHP 8.2 broke something again :disappointed:

PR #327, opened 2022-10-20
3.0 - Add php type declarations

othercorey:
This enables the cake5 phpcs branch.

Various self, $this and ChronosInterface return types were synchronized between the interface and traits/classes.

### DIFF
--- a/src/Traits/FactoryTrait.php
+++ b/src/Traits/FactoryTrait.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
  */
 namespace Cake\Chronos\Traits;
 
-use Cake\Chronos\ChronosInterface;
 use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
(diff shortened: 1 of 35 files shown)


src/Traits/FactoryTrait.php
```diff
@@ -53,12 +52,14 @@ public static function instance(DateTimeInterface $dt): ChronosInterface
      * ChronosInterface::parse('Monday next week')->fn() rather than
      * (new Chronos('Monday next week'))->fn()
      *
-     * @param \DateTimeInterface|string|int $time The strtotime compatible string to parse
+     * @param \DateTimeInterface|string|int|null $time The strtotime compatible string to parse
```
othercorey, line 55: Added to match unit tests which pass null to the constructor.

src/Traits/MagicPropertyTrait.php
```diff
@@ -55,10 +56,10 @@ trait MagicPropertyTrait
      * Get a part of the ChronosInterface object
      *
      * @param string $name The property name to read.
-     * @return string|int|bool|\DateTimeZone The property value.
+     * @return \DateTimeZone|string|float|int|bool The property value.
```
othercorey, line 59: Added `float` for `offsetHours`.

PR #338, opened 2022-10-26
Rename method args for consistency and better readability.

ADmad:


src/Chronos.php
```diff
@@ -88,18 +88,20 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
...
      *
      * @param \DateTimeInterface|string|int|null $time Fixed or relative time
-     * @param \DateTimeZone|string|null $tz The timezone for the instance
+     * @param \DateTimeZone|string|null $timezone The timezone for the instance
      */
-    public function __construct(DateTimeInterface|string|int|null $time = 'now', DateTimeZone|string|null $tz = null)
-    {
+    public function __construct(
+        DateTimeInterface|string|int|null $time = 'now',
+        DateTimeZone|string|null $timezone = null
```
othercorey, line 95: Since time zone is 2 words, I think we should follow php use `timeZone` to be accurate.

othercorey: Hmm, php names the class TimeZone but the constructor parameter for DateTime is all lower case `timezone`? Unfortunate.

We won't be extending DateTime though so ...

ADmad: Yes, PHP capitalizes the `z` only for the class name. For arguments (`$timezone`) and methods names (`setTimezone()`) it's lowercased. I would prefer we keep the same even if we are not going to extend `DateTimeImmutable`.

othercorey: yea, let's follow it.

othercorey (approved):

PR #340, opened 2022-10-28
Simplify ChronosInterval.

ADmad:
It now only provides a static method for creating it's instances and __toString() to get the ISO 8601 interval string.

All methods which manipulate the properties of the extended DateInterval have been dropped. The PHP manual states that they should be considered readable only. An exception is made only for DateInterval::$f as the constructor doesn't support specifying microseconds through the interval string.

### DIFF
--- a/src/ChronosInterval.php
+++ b/src/ChronosInterval.php
@@ -14,43 +14,12 @@ declare(strict_types=1);
  */
 namespace Cake\Chronos;
 
-use BadMethodCallException;
 use DateInterval;
-use InvalidArgumentException;
 
 /**
  * A simple API extension for DateInterval.
- * The implementation provides helpers to handle weeks but only days are saved.
- * Weeks are calculated based on the total days of the current instance.
  *
- * @property int $years Total years of the current interval.
- * @property int $months Total months of the current interval.
- * @property int $weeks Total weeks of the current interval calculated from the days.
- * @property int $dayz Total days of the current interval (weeks * 7 + days).
- * @property int $hours Total hours of the current interval.
- * @property int $minutes Total minutes of the current interval.
- * @property int $seconds Total seconds of the current interval.
- * @property int $microseconds Total microseconds of the current interval.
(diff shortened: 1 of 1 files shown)


ADmad:
I will update the tests once the change is approved in principal.

src/ChronosInterval.php
```diff
@@ -166,298 +74,60 @@ public function __construct(
...
+        ?int $years = null,
         ?int $months = null,
         ?int $weeks = null,
         ?int $days = null,
         ?int $hours = null,
         ?int $minutes = null,
         ?int $seconds = null,
-        ?int $microseconds = null
+        ?int $microseconds = null,
     ): static {
```
markstory, line 85: Nice! I'm glad to see we're keeping a method that provides this more ergonomic signature over the hard to remember interval string format that PHP uses.

src/ChronosInterval.php
```diff
@@ -166,298 +74,60 @@ public function __construct(
...
-
-    /**
-     * Set a part of the ChronosInterval object
-     *
-     * @param string $name The property to augment.
-     * @param int|false $val The value to change.
-     * @return void
-     * @throws \InvalidArgumentException
-     */
-    public function __set(string $name, int|false $val): void
```
markstory, line 315: Without these mutation methods how would users modify or adjust an interval once it has been made? Should we also have a `modifyWith()` method that has a similar interface to `create()` that returns a new updated interval?

ADmad: Are you aware of the use cases for needing to modify an interval? `DateInterval` doesn't  provided any APi to do so. The only way to modify an interval after creation is manipulating it properties directly and the PHP manuals warns against doing so and treating the ~~methods~~ properties as read only.

Me and @othercorey discussed this a bit more and think just having a `Chronos::createInternal()` which reruns a `DateInterval` would suffice and we can drop `ChronosInterval` all together.

markstory: I don't have use cases in mind for modifying an interval. We can always add more API surface in the future :smile:

PR #341, opened 2022-10-31
Add DateIntervalFactory.

ADmad:
This is a better/simpler alternative to ChronosInterval which extends DateInterval.

Alternative to #340. We can simply drop ChronosInterval, it's unnecessary complexity.


othercorey:
We could add the factory methods to Chronos `Chronos::createInterval()`.

ADmad:
Okay, that works too. Will make a new PR.

PR #342, opened 2022-11-01
Remove not needed legacy API method.

dereuromark:
If we now fix up the API and method naming, I suggest renaming this as the current one, removes the acronym here.
Is already a duplicate somewhat in naming anyway ("TimestampUTC" => TimestampUniversalTimeCoordinated)

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -547,7 +547,7 @@ class Chronos
      */
     public static function maxValue(): static
     {
-        return static::createFromTimestampUTC(PHP_INT_MAX);
+        return static::createFromTimestamp(PHP_INT_MAX);
     }
 
     /**
@@ -559,7 +559,7 @@ class Chronos
     {
         $max = PHP_INT_SIZE === 4 ? PHP_INT_MAX : PHP_INT_MAX / 10;
 
-        return static::createFromTimestampUTC(~$max);
+        return static::createFromTimestamp(~$max);
     }
 
     /**
@@ -751,17 +751,6 @@ class Chronos
     public static function createFromTimestamp(int $timestamp, DateTimeZone|string|null $timezone = null): static
     {
         return static::now($timezone)->setTimestamp($timestamp);
-    }
-
-    /**
-     * Create an instance from an UTC timestamp
-     *
-     * @param int $timestamp The UTC timestamp to create an instance from.
-     * @return static
-     */
-    public static function createFromTimestampUTC(int $timestamp): static
-    {
-        return new static($timestamp);
     }
 
     /**
--- a/tests/Benchmark/ConstructBench.php
+++ b/tests/Benchmark/ConstructBench.php
@@ -170,6 +170,6 @@ class ConstructBench
     public function benchFromTimestampUTC($params)
     {
         $class = $params['class'];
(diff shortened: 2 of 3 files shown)


ADmad:
This is unnecessarily verbose. UTC is a very well known acronym, there's no need to expand it.

dereuromark:
The wording would then be createFromTimestampUtc however given that acronyms are not all uppercase in method parts.
I still dont really like the UTC part in there much.

dereuromark:
Also `createFromTimestamp()` itself would probably be enough to clarify the intent. They are always "universal" in that scope, no?

ADmad:
Yes, `createFromTimestamp()` would suffice. Timestamps are always in UTC.

dereuromark:
Ah, apparently there is already a unused? method in there with that name for a timezone based timestamp (if that makes sense):
```php
    /**
     * Create a ChronosInterface instance from a timestamp
     *
     * @param int $timestamp The timestamp to create an instance from.
     * @param \DateTimeZone|string|null $timezone The DateTimeZone object or timezone name the new instance should use.
     * @return static
     */
    public static function createFromTimestamp(int $timestamp, DateTimeZone|string|null $timezone = null): static
    {
        return static::now($timezone)->setTimestamp($timestamp);
    }

    /**
     * Creates a ChronosInterface instance from an UTC timestamp.
     *
     * @param int $timestamp The UTC timestamp to create an instance from.
     * @return static
     */
    public static function createFromUniversalTimestamp(int $timestamp): static
    {
        return new static($timestamp);
    }
```
If we remove that one, the method name becomes available. Or the first one should maybe be named `createFromTimestampWithTimezone()`.

ADmad:
Do we even need 2 methods? We could have just 1.

```php
    public static function createFromTimestamp(int $timestamp, DateTimeZone|string|null $timezone = null): static
    {
        if ($timezone) {
            return static::now($timezone)->setTimestamp($timestamp);
        }

        return new static($timestamp);
    }
```

othercorey:
The `createFromTimestamp()` method keeps the the user-defined timezone. This is what happens when you call `setTimestamp()` on a `DateTime` instance.

othercorey:
This would be a lot less confusing since it can be documented in one place.

dereuromark:
Seems like the suggested changes are not backed by tests (yet).

ADmad:
Hmm.. The proposed change changes how `createFromTimestamp()` works. Currently the method uses the current timezone if none is specified.

dereuromark:
Feel free to ammend the PR as you see fit. I currently dont know what we can do here if we want to keep the single method.

dereuromark:
How do we want to proceed here now?

markstory:
I think what you have here is good. We don't need a timezone on the simple integer value as the `Chronos` constructor will append any timestamp values with `@` making them UTC implicitly.

dereuromark:
you mean we can remove this method?

markstory:
> you mean we can remove this method?

yes, the `createFromTimestampUTC` method seems a bit redundant to me.

dereuromark:
@markstory Can we remove that method now in RC still?

dereuromark:
OK, I rebased.
So we can remove it then?

PR #343, opened 2022-11-02
 Switch human diff to months when interval is 2+ months

othercorey:
refs https://github.com/cakephp/chronos/pull/323


othercorey:
@markstory Should we manually calculate the month difference or rely on PHP here? We should not jump from 7 weeks to 1 month.

markstory:
Hrm. 7 weeks to 1 month does seem wrong. Could we do it with the number of days instead?

othercorey:
> Hrm. 7 weeks to 1 month does seem wrong. Could we do it with the number of days instead?

Decided to switch to months only when the interval is showing 2+ months. That is after 8 weeks but depends on the day of the month on how many days. `addMonth(2)` and `subMonths(2)` should trigger it.

PR #344, opened 2022-11-04
Start deprecations for 2.next

markstory:
The tests are currently failing, but I was conflicted on how to proceed.

* Should we add a method like `deprecated(Closure)` to chronos' test library?
* Should we copy the `deprecationWarning` helper from Cake to chronos as well?

### DIFF
--- a/src/Traits/ModifierTrait.php
+++ b/src/Traits/ModifierTrait.php
@@ -494,7 +494,10 @@ trait ModifierTrait
      */
     public function addMonthWithOverflow(int $value = 1): ChronosInterface
     {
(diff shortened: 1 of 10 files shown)


src/Traits/ModifierTrait.php
```diff
@@ -296,6 +296,8 @@ public function addYears(int $value): ChronosInterface
      */
     public function addYear(int $value = 1): ChronosInterface
     {
+        trigger_error('2.4 - addYear() is deprecated. Use addYears() instead.', E_USER_DEPRECATED);
```
othercorey: Let's us a similar format to cake core. `Since 2.4.0, ...`

markstory: Sounds good.

Issue #79, opened 2016-02-03
Chronos\ChronosInterface usefulness


nyamsprod:
``` php
function doSomething(\Cake\Chronos\ChronosInterface $chronos)
{
    return $chronos->year(2015);
}
```

If a `Cake\Chronos\Chronos` object is given to the function returns a new `Cake\Chronos\Chronos` because this object is immutable

If a `Cake\Chronos\MutableDateTime` object is given to the function will modify the object and return a reference to it because it is mutable

So you can not safely typehint against the `Cake\Chronos\ChronosInterface` and be safe regarding
- the input being changed or not
- the output being a new object or a reference to input

This means that you should typehint against a concrete implementation which reduce the interface usefulness

From a developer point this is a design issue. Of note, PHP's `DateTimeInterface` avoid this problem 
by only specifying methods which do not affect the object internal state.

josegonzalez:
As I mentioned before:

> ChronosInterface exists so we can typecheck as to whether or not the object coming in is a type of Date/DateTime object from the Chronos package, and we would use a more specific typehint where necessary.

Maybe we should have a `MutableChronosInterface` and `ImmutableChronosInterface` as well, so you can check both that the object inherits from the Chronos package as well as it's mutability?

nyamsprod:
A simpler solution would be to copy `DateTimeInterface` (ie: to only put non modifying methods in the common interface). This way you get to safely typehint against it;

Each class modifying method behaves differently even if they share the same name so I see no added value to have them in the interface. (ie: `Chronos::add` != `MutableDateTime::add` != `Date::add` != `MutableDate:add`)

markstory:
The main goal with the ChronosInterface was to allow separation between native DateTime objects and enriched ones like @josegonzalez mentioned.

I can see how the current interface hints don't allow you to ensure immutability though. What do you think of having a few more 'marker' interfaces:

* `Cake\Chronos\ImmutableInterface` - For the immutable variants.
* `Cake\Chronos\MutableInterface` - For the mutable variants.
* `Cake\Chronos\DateInterface` - For the date objects that lack time.

I think this gives enough typehints to give what you're looking for, but I might be missing something.

nyamsprod:
I think that:

- `Cake\Chronos\ChronosInterface` should only contains **non modifying methods**
- `Cake\Chronos\ChronosInterface` should not extend `DateTimeInterface`
[...]

PR #345, opened 2022-11-04
Wrap DateTimeImmutable instead of extending it

othercorey:
closes https://github.com/cakephp/chronos/issues/79


src/Chronos.php
```diff
@@ -182,6 +196,112 @@ public static function hasTestNow(): bool
...
+     *
+     * @param string $modifier Date modifier
+     * @return static
+     * @throws \InvalidArgumentException
+     */
+    public function modify(string $modifier): static
+    {
+        $native = $this->native->modify($modifier);
+        if ($native === false) {
+            throw new InvalidArgumentException('Unable to modify date using: ' . $modifier);
```
othercorey, line 855: This throws an exception on error now.

src/Traits/FrozenTimeTrait.php
```diff
@@ -132,26 +145,35 @@ public function setTimezone(DateTimeZone|string $value): static
...
-            return $this;
+        if (preg_match('/hour|minute|second/', $modifier)) {
+            return clone $this;
         }
-        $new = parent::modify($relative);
+
+        $new = clone $this;
+        $new->native = $new->native->modify($modifier);
+        if ($new->native === false) {
+            throw new InvalidArgumentException('Unable to modify date using: ' . $modifier);
```
othercorey, line 146: This throws an exception on error now

tests/TestCase/Date/TimezoneTest.php
```diff
@@ -27,10 +27,8 @@ class TimezoneTest extends TestCase
      */
     public function testNoopOnTimezoneChange()
     {
-        $tz = new DateTimeZone('Pacific/Honolulu');
         $date = new Date('2015-01-01');
-        $new = $date->setTimezone($tz);
```
othercorey, line 32: `Date::setTimezone()` was incorrectly returning `$this` instead of a new instance.

src/Chronos.php
```diff
@@ -81,46 +83,59 @@ class Chronos extends DateTimeImmutable implements ChronosInterface
...
+        if (is_int($time)) {
+            return new DateTimeImmutable("@{$time}");
         }
 
         if ($timezone !== null) {
             $timezone = $timezone instanceof DateTimeZone ? $timezone : new DateTimeZone($timezone);
         }
 
-        if ($time instanceof DateTimeInterface) {
+        if (is_object($time)) {
```
ADmad, line 251: The `Date` class will also have a timezone?

src/Chronos.php
```diff
@@ -182,6 +196,112 @@ public static function hasTestNow(): bool
...
+    /**
+     * Sets the time.
+     *
+     * @param int $hours Hours of the time
+     * @param int $minutes Minutes of the time
+     * @param int $seconds Seconds of the time
+     * @param int $microseconds Microseconds of the time
+     * @return static
+     */
+    public function setTime(int $hours, int $minutes, int $seconds = 0, int $microseconds = 0): static
```
ADmad, line 835: Can we have all arguments default to 0? That would allow one to do for e.g. `->setTime(minutes: 1)`.

src/Chronos.php
```diff
@@ -182,6 +196,112 @@ public static function hasTestNow(): bool
...
+        return $new;
+    }
+
+    /**
+     * Creates a new instance with date modified according to DateTimeImmutable::modifier().
+     *
+     * @param string $modifier Date modifier
+     * @return static
+     * @throws \InvalidArgumentException
+     */
```
ADmad, line 850: A `@see` or `@link` tag to the PHP manual for the modifier string format would be nice.

src/Chronos.php
```diff
@@ -182,6 +196,112 @@ public static function hasTestNow(): bool
...
+    }
+
+    /**
+     * Return time zone set for this instance.
+     *
+     * @return \DateTimeZone|false
+     */
+    public function getTimezone(): DateTimeZone|false
+    {
+        return $this->native->getTimezone();
```
ADmad: Are you aware in what case this would return `false`? Manual just says "...false in case of failure".

src/Chronos.php
othercorey: It currently is set to the default time zone.

src/Chronos.php
othercorey: We can. This matched DateTimeImmutable::setTime(). I'm not sure why it was defaulting to null and not 0.

src/Chronos.php
othercorey: I could not figure it out. If you look at the PHP source, it is false if some internal timelib flag "is local time" is set. I have no idea how that happens.

ADmad: I would change the return type for our method to just `DateTimeZone` and throw an exception if the native class returns false.

markstory: Being more strict sounds good to me too.

othercorey:
Dropped ChronosInterface and moved all constants to Chronos. Moved all static configs to Chronos since they should be the same for all instances.

Did not move **all** trait functions into Chronos and ChronosDate only the ones that seem to require type changes. Hopefully that makes it easier to review.

composer.json
```diff
@@ -50,10 +50,12 @@
...
         "cs-fix": "phpcbf --colors --parallel=16 -p src/ tests/",
         "phpstan": "phpstan.phar analyse",
+        "psalm": "psalm.phar --show-info=false",
         "stan": [
-            "@phpstan"
+            "@phpstan",
+            "@psalm"
         ],
-        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.8.0 && mv composer.backup composer.json",
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:~1.8.0 psalm/phar:~5.0-beta && mv composer.backup composer.json",
```
ADmad, line 58: We should just drop this now that we use the common workflows for CI and the versions used here will keep getting out of sync.

src/Traits/FrozenTimeTrait.php
```diff
@@ -33,13 +35,15 @@ trait FrozenTimeTrait
...
-     * @param \DateTime|\DateTimeImmutable|string|int|null $time The input time. Integer values will be assumed
+     * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosDate|\DateTime|\DateTimeImmutable|string|int|null $time The input time. Integer values will be assumed
      *   to be in UTC. The 'now' and '' values will use the current local time.
      * @param \DateTimeZone|null $timezone The timezone in which the date is taken
      * @return string The date component of $time.
      */
-    protected function stripTime(DateTime|DateTimeImmutable|string|int|null $time, ?DateTimeZone $timezone): string
-    {
+    protected function stripTime(
+        Chronos|ChronosDate|DateTime|DateTimeImmutable|string|int|null $time,
```
ADmad:
```suggestion
        Chronos|ChronosDate|DateTimeInterface|string|int|null $time,
```

src/Traits/FrozenTimeTrait.php
```diff
@@ -33,13 +35,15 @@ trait FrozenTimeTrait
      *
      * Used to ensure constructed objects always lack time.
      *
-     * @param \DateTime|\DateTimeImmutable|string|int|null $time The input time. Integer values will be assumed
+     * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosDate|\DateTime|\DateTimeImmutable|string|int|null $time The input time. Integer values will be assumed
```
ADmad:
```suggestion
     * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosDate|\DateTimeInterface|string|int|null $time The input time. Integer values will be assumed
```

composer.json
othercorey: I just need something to install locally.

tests/TestCase/Date/DiffTest.php
```diff
@@ -0,0 +1,375 @@
...
+
+    /* Does not include January 31 since DatePeriod does not include the end date by default
+    public function testDiffInDaysFilteredPositiveWithMutated()
+    {
+        $dt = ChronosDate::createFromDate(2000, 1, 1);
+        $this->assertSame(5, $dt->diffInDaysFiltered(function ($date) {
+            return $date->dayOfWeek === 1;
+        }, $dt->endOfMonth()));
+    }
+    */
```
othercorey, line 124: These disabled tests just happens to pass with Chronos because we use the current time instead of setting to 00:00:00. DatePeriod does not include the end date and the option to enable that was added in PHP 8.2.

othercorey: Think I can work around this in 8.1 by adding time to the end date.

src/Chronos.php
```diff
@@ -183,6 +297,46 @@ public static function hasTestNow(): bool
...
+    public static function setWeekendDays(array $days): void
+    {
+        static::$weekendDays = $days;
+    }
+
+    /**
+     * Get the difference formatter instance or overwrite the current one.
+     *
+     * @param \Cake\Chronos\DifferenceFormatterInterface|null $formatter The formatter instance when setting.
+     * @return \Cake\Chronos\DifferenceFormatterInterface The formatter instance.
```
markstory, line 393: We could have a cake adapter for this interface that handles the translation aspects. That could be a way to resolve the 'how to align the cake and chronos feature set'. We align the formatting in the new version of Chronos, and then cake would provide the i18n integration.

src/Chronos.php
```diff
@@ -247,6 +401,552 @@ public static function createInterval(
...
+    /**
+     * Returns the difference between this instance and target.
+     *
+     * @param \Cake\Chronos\Chronos|\DateTimeInterface $target Target instance
+     * @param bool $absolute Whether the interval is forced to be positive
+     * @return \DateInterval
+     */
+    public function diff(Chronos|DateTimeInterface $target, bool $absolute = false): DateInterval
+    {
+        $target = $target instanceof DateTimeInterface ? $target : $target->native;
```
markstory: Love this tech of using PHP's scoping rules to access other objects :clap:

src/Chronos.php
```diff
@@ -247,6 +401,552 @@ public static function createInterval(
...
+
+    /**
+     * Determines if the instance is greater (after) than another
+     *
+     * @param \Cake\Chronos\Chronos $other The instance to compare with.
+     * @return bool
+     */
+    public function greaterThan(Chronos $other): bool
+    {
+        return $this > $other;
```
markstory:
```suggestion
        return $this->native > $other->native;
```

src/ChronosDate.php
```diff
@@ -115,8 +142,416 @@ public function __construct(
...
+
+    /**
+     * Return time zone set for this instance.
+     *
+     * @return \DateTimeZone|false
+     */
+    public function getTimezone(): DateTimeZone|false
+    {
+        return $this->native->getTimezone();
+    }
```
markstory: I don't think dates should have Timezones. This was a mistake in the previous design that we could correct now. I'm happy to do that as a follow up pull request if you'd prefer.

tests/TestCase/Date/DiffTest.php
markstory: Having date operations behave differently based on the time of day they are run is unexpected. Date operations should behave like calendar dates do.

I'm happy to do a follow up pull request to improve how this works.

src/ChronosDate.php
othercorey: I was just re-creating the api from the interface initially, but we should drop this and see what needs to be fixe in cake core.

othercorey:
Dropped time zone from ChronosDate and any functions affected by time zone. Dropped creating `ChronosDate()` from now.

Dropped:

* set/getTimezone
* getOffset
* set/getTimestamp
* now
* instance (same as parse)
* today/tomorrow
* isFuture/isPast/isNextWeek
* Related magic getters

Users can get the current date through Chronos: `new ChronosDate(Chronos::now('America/Toronto'))`

Moved many functions into Chronos and Chronos date to allow changes in implementation.

othercorey:
I also dropped `createFromDate()` since `create()` is `createFromDate()` for `ChronosDate`. I don't know if we want to keep that duplicate.

docs/ja/index.rst
```diff
@@ -286,7 +286,7 @@ Chronos は、出力した日時オブジェクトを表示するための多く
 
     Chronos::setTestNow(Chronos::now());
     MutableDateTime::setTestNow(MutableDateTime::now());
-    ChronosDate::setTestNow(ChronosDate::now());
+    ChronosDate::setTestNow(ChronosDate::parse(Chronos::now()));
```
markstory, line 289: I don't love this API but its something we can simplify after this merges.

markstory (approved):
Looks good to me. Lets merge it.

docs/ja/index.rst
othercorey: This feels like some sort of search and replace error. It should just be now().

PR #346, opened 2022-11-04
Drop ChronosInterval and add Chronos::createInterval().

ADmad:
Most methods of ChronosInterval manipulated the properties of the extended DateInterval. The PHP manual states that the properties should be considered readable only. Extending PHP's datetime related classes also often causes issues in new PHP minors. Hence the class has been dropped to avoid problems in the future.


othercorey (approved):

src/Chronos.php
```diff
@@ -182,6 +183,66 @@ public static function hasTestNow(): bool
...
+     * @param int|null $months The month to use.
+     * @param int|null $weeks The week to use.
+     * @param int|null $days The day to use.
+     * @param int|null $hours The hours to use.
+     * @param int|null $minutes The minutes to use.
+     * @param int|null $seconds The seconds to use.
+     * @param int|null $microseconds The microseconds to use.
+     * @return \DateInterval
+     */
+    public static function createInterval(
```
markstory, line 199: This is something we can backport reasonably easy. We could also deprecate the entire `ChronosInterval` as well.

ADmad: Yes, this should be easy to backport.

PR #348, opened 2022-11-08
2.next - Rename Date to ChronosDate. Deprecate MutableDate and MutableDateTime.

othercorey:


src/ChronosDate.php
```diff
@@ -0,0 +1,144 @@
...
+ * @property-read int $quarter the quarter of this instance, 1 - 4
+ * @property-read int $offset the timezone offset in seconds from UTC
+ * @property-read int $offsetHours the timezone offset in hours from UTC
+ * @property-read bool $dst daylight savings time indicator, true if DST, false otherwise
+ * @property-read bool $local checks if the timezone is local, true if local, false otherwise
+ * @property-read bool $utc checks if the timezone is UTC, true if UTC, false otherwise
+ * @property-read string $timezoneName
+ * @property-read string $tzName
+ */
+class ChronosDate extends DateTimeImmutable implements ChronosInterface
```
markstory, line 54: I'm assuming that this is where the breaking changes to `Date` should be made? I've been thinking through how that could work.

othercorey: This is just renaming it so any changes should be in `ChronosDate` after merged.

jurchiks:
Why did you rename this class? `Date` was the PERFECT name for it!

PR #349, opened 2022-11-09
2.next Backport createInterval()

markstory:
Backport createInterval() and update diffFiltered() to be compatible with `DateInterval`.

The `createInterval()` method is pretty awkward to use without named parameters. But I don't think there is anything we can do there other than to encourage users to update to PHP8.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,6 @@ want to migrate, we could use the following to update files:
 
 ```
 # Replace imports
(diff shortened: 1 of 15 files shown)


ADmad:
> The createInterval() method is pretty awkward to use without named parameters.

Yes the convenience only exists if one uses named params :)

PR #351, opened 2022-11-21
Re-add ChronosDate::now() and other time-relative helpers

othercorey:
refs https://github.com/cakephp/chronos/pull/345

### DIFF
--- a/src/ChronosDate.php
+++ b/src/ChronosDate.php
@@ -18,6 +18,7 @@ use DateInterval;
 use DatePeriod;
 use DateTimeImmutable;
 use DateTimeInterface;
+use DateTimeZone;
 use InvalidArgumentException;
 
 /**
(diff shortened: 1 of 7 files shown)


tests/TestCase/Date/ConstructTest.php
```diff
@@ -97,6 +98,79 @@ public function testConstructWithRelative()
         $this->assertSame('2001-01-08', $c->format('Y-m-d'));
     }
 
+    public function testConstructWithLocalTimezone(): void
```
othercorey, line 101: These tests were copied from Chronos tests and modified to fit ChronosDate.

tests/TestCase/Date/IsTest.php
```diff
@@ -40,6 +40,162 @@ public function testIsWeekendFalse()
         $this->assertFalse(ChronosDate::create(2012, 1, 2)->isWeekend());
     }
 
+    public function testIsYesterdayTrue()
```
othercorey, line 43: These tests were copied from Chronos and modified to fit ChronosDate.

tests/TestCase/Date/IsTest.php
```diff
@@ -40,6 +40,162 @@ public function testIsWeekendFalse()
...
+    }
+
+    public function isTodayFalseWithTimezone()
+    {
+        date_default_timezone_set('Pacific/Kiritimati');
+        $samoaTimezone = new DateTimeZone('Pacific/Samoa');
+
+        // Pacific/Samoa -11:00 is used intead of local timezone +14:00
+        $this->assertFalse(ChronosDate::now()->isToday($samoaTimezone));
+        $this->assertTrue(ChronosDate::now()->isToday('Pacific/Kiritimati'));
```
othercorey, line 80: Added test to confirm date is taken from the correct time zone when comparing.

src/ChronosDate.php
```diff
@@ -965,6 +1023,127 @@ public function isWeekend(): bool
         return in_array($this->dayOfWeek, Chronos::getWeekendDays(), true);
     }
 
+    /**
+     * Determines if the instance is yesterday
+     *
+     * @param \DateTimeZone|string|null $timezone Time zone to use for now.
+     * @return bool
+     */
+    public function isYesterday(DateTimeZone|string|null $timezone = null): bool
```
othercorey, line 1032: These helpers now have a time zone parameter to allow comparing to the correct `now` but defaults to server time.

src/ChronosDate.php
```diff
@@ -90,35 +91,92 @@ class ChronosDate
...
      *
      * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosDate|\DateTimeInterface|string $time Fixed or relative time
+     * @param \DateTimeZone|string|null $timezone The time zone used for 'now'
      */
-    public function __construct(Chronos|ChronosDate|DateTimeInterface|string $time)
-    {
-        $this->native = $this->createNative($time);
+    public function __construct(
+        Chronos|ChronosDate|DateTimeInterface|string $time = 'now',
+        DateTimeZone|string|null $timezone = null
```
othercorey, line 98: Added back `now` default but did not re-add `null`.  Added time zone parameter back since this is the equivalent to `now()`.

ADmad:
Looking at the current failures in the Cake core we might probably need to add `getTimezone()/setTimezone()` to `ChronosDate` or curtail where date objects can be used in various comparison / formatting functions.

othercorey:
> Looking at the current failures in the Cake core we might probably need to add `getTimezone()/setTimezone()` to `ChronosDate` or curtail where date objects can be used in various comparison / formatting functions.

I have a PR for cakephp that cleans most of that up.

src/ChronosDate.php
```diff
@@ -90,35 +91,92 @@ class ChronosDate
...
      * @return \DateTimeImmutable
      */
-    protected function createNative(Chronos|ChronosDate|DateTimeInterface|string|int|null $time): DateTimeImmutable
-    {
+    protected function createNative(
+        Chronos|ChronosDate|DateTimeInterface|string $time,
+        DateTimeZone|string|null $timezone
+    ): DateTimeImmutable {
+        if (!is_string($time)) {
+            return new DateTimeImmutable($time->format('Y-m-d 00:00:00'));
```
markstory, line 115: Should this use the timezone parameter?

othercorey: No, the time zone is only for interpreting the time string or now.

markstory: Ok. I was concerned that we'll be shifting timezones to server default from the provided datetime. Thinking about it more though what you have is correct as we're preserving the date aspect and timezones are only relevant when figuring out relative concepts like today/yesterday/tomorrow.

PR #352, opened 2022-11-26
3.x Small ChronosDate cleanup

markstory:


tests/TestCase/Date/DiffTest.php
```diff
@@ -113,15 +113,16 @@ public function testDiffInDaysVsDefaultNow()
...
 
-    /* Does not include January 31 since DatePeriod does not include the end date by default
+    /**
+     * DateInterval does not include the end date by default.
+     */
     public function testDiffInDaysFilteredPositiveWithMutated()
     {
         $dt = ChronosDate::create(2000, 1, 1);
-        $this->assertSame(5, $dt->diffInDaysFiltered(function ($date) {
+        $this->assertSame(4, $dt->diffInDaysFiltered(function ($date) {
```
othercorey: I think we should try modifying the DatePeriod used by adding 1 day so it always includes the end date. If we switch to PHP 8.2 before release, we can use the flag. This makes the API correct vs just adjusting the edge case unit test.

PR #355, opened 2022-11-27
Expand docs a bit for new Date objects

markstory:


docs/en/index.rst
```diff
@@ -89,8 +89,8 @@ Date Objects
 
 PHP only provides a single DateTime object. Representing calendar dates can be
 a bit awkward with this class as it includes timezones, and time components that
-don't really belong in the concept of a 'day'. Chronos provides a ``Date``
-object that allows you to represent dates. The time and timezone for these
+don't really belong in the concept of a 'day'. Chronos provides ``ChronosDate``
```
othercorey: I want to re-write the beginning of this sentence to:

PHP only provides a date and time class - ``DateTime``. Representing calendar dates can be
a bit awkward with ``DateTime`` as it includes time and time zones that aren't part of a "date".

markstory: Good idea. I reworded this bit based on your suggestion.

PR #356, opened 2022-12-07
3.x Clean up traits and remove methods from Date

markstory:
Inline most of the traits as they only have one usage each. The FormattingTrait was moved up a directory as having a directory with only one file feels a bit silly.

I've also marked several classes as internal because I don't think that userland code should be relying on these classes and logic.


src/ChronosDate.php
```diff
@@ -272,6 +270,83 @@ public static function diffFormatter(?DifferenceFormatterInterface $formatter =
...
+     * @param int $value The timestamp value to set.
+     * @return static
+     */
+    public function setTimestamp(int $value): static
+    {
+        $new = clone $this;
+        $new->native = $new->native->setTimestamp($value)->setTime(0, 0, 0);
+
+        return $new;
+    }
```
othercorey: ChronosDate shouldn't have this method.

src/ChronosDate.php
```diff
@@ -272,6 +270,83 @@ public static function diffFormatter(?DifferenceFormatterInterface $formatter =
...
+     * Subtract an Interval from a Date.
+     *
+     * Any changes to the time will be ignored and reset to 00:00:00
+     *
+     * @param \DateInterval $interval The interval to modify this date by.
+     * @return static A modified Date instance
+     */
+    public function sub(DateInterval $interval): static
+    {
+        $new = clone $this;
```
othercorey, line 305: Should probably assert that hours, minutes second and microseconds aren't set.

markstory: Isn't resetting the time components enough to prevent the time from being retained? Or are you thinking of throwing an error if the interval includes hours/minutes/seconds/microseconds?

othercorey: I was thinking an assert if the interval has those set. The idea is to make users aware of input that isn't kept. However, if you think users will use intervals created from Chronos and not ChronosDate objects, then we can ignore it.

markstory: :+1: Good idea, I've made that change and also updated `modify()` so that it also disallows modification of time components for consistency.

PR #363, opened 2022-12-14
2.5 - Backport date

markstory:
There are a few things going on here as I started small but wanted to
stop before the diff got too large.

- Added deprecations for modifing dates by their time components.
- I've disallowed diffing dissimilar types with a runtime check. The
  goal of this was to approximate the type errors that will be present
  in 3.x.
- Add deprecations for short comparison methods.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,6 +1,11 @@
 parameters:
 	ignoreErrors:
 		-
+			message: "#^Strict comparison using \\=\\=\\= between class\\-string\\<Cake\\\\Chronos\\\\Chronos\\> and 'Cake\\\\\\\\Chronos\\\\\\\\ChronosDate' will always evaluate to false\\.$#"
+			count: 2
+			path: src/Chronos.php
+
+		-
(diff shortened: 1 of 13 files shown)


othercorey:
Look ok.

phpstan-baseline.neon
```diff
@@ -1,5 +1,10 @@
 parameters:
 	ignoreErrors:
+		-
+			message: "#^Strict comparison using \\=\\=\\= between class\\-string\\<Cake\\\\Chronos\\\\Chronos\\> and 'Cake\\\\\\\\Chronos\\\\\\\\ChronosDate' will always evaluate to false\\.$#"
+			count: 2
+			path: src/Chronos.php
```
othercorey, line 6: Whatever code is causing this error seems incorrect. These will never evaluate correctly.

markstory: These warnings were coming from the traits, and are related to the trait method having an impossible condition in some of the implementations.

othercorey: Ok, gotcha.

PR #367, opened 2022-12-17
2.5 - More deprecations

markstory:
* Deprecate `toMutable()`
* Deprecate factory methods that won't exist on ChronosDate in the future.


othercorey:
Looks ok

PR #373, opened 2023-01-12
Fix up getTimezone()

dereuromark:
Before we release RCs, we need to raise PHPStan to level 8
There are quite a few bugs and issues around still regarding false and nullable etc.

`|false` in many cases is a code smell, making things outside the API more complicated.
Here we dont need it, but also we cannot use it, since e.g.

    Chronos::now($first->getTimezone());
    
Expects `DateTimeZone|string|null $timezone = null`

So better to clean up the API to the outside.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -1000,11 +1000,11 @@ class Chronos
     /**
      * Return time zone set for this instance.
      *
-     * @return \DateTimeZone|false
+     * @return \DateTimeZone|null
      */
-    public function getTimezone(): DateTimeZone|false
+    public function getTimezone(): ?DateTimeZone
     {
-        return $this->native->getTimezone();
+        return $this->native->getTimezone() ?: null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ADmad:
> |false in many cases is a code smell..

I don't see how that's an issue anymore in PHP 8.1+ since `false` can now be used as a union type.

dereuromark:
It does not fit to a clean API in general
Most `|false` inside PHP are also just legacy, no need to put this into a new major API for years to come.
It also requires more case checking within the methods, as the code above shows. For no reason IMO.

ADmad:
For the particular e.g. you provided returning `null` instead of `false` does make sense but in general I would prefer the API matches the core PHP interfaces/classes.

dereuromark:
Not a big fan of those dirty methods, but fair enough I guess.

markstory:
I think this change is a good one. Maintaining the `DateTime|false` just because that's how PHP works doesn't feel like the right choice. One of the reasons we stopped extending PHP's object was so that we could simplify and improve return values/parameter types.

PR #378, opened 2023-01-26
PHPStan fixes, set level to 8 now

dereuromark:
Also solved the remaining ones of level 8:
```
 ------ ---------------------------------------------------------------------------------------------------------------------- 
  Line   Chronos.php                                                                                                           
 ------ ---------------------------------------------------------------------------------------------------------------------- 
  282    Parameter #1 $modifier of method Cake\Chronos\Chronos::modify() expects string, string|null given.                    
  2543   Method Cake\Chronos\Chronos::__get() should return bool|DateTimeZone|float|int|string but returns DateTimeZone|null.  
  2549   Cannot call method getName() on DateTimeZone|null.                                                                    
 ------ ---------------------------------------------------------------------------------------------------------------------- 


                                                                                                                        
 [ERROR] Found 3 errors 
```

Should be all green now.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -268,13 +268,13 @@ class Chronos
         }
 
         $relative = static::hasRelativeKeywords($time);
-        if (!empty($time) && $time !== 'now' && !$relative) {
+        if ($time && $time !== 'now' && !$relative) {
             return new DateTimeImmutable($time, $timezone);
         }
 
         $testNow = clone $testNow;
-        $relativetime = self::isTimeExpression($time);
-        if (!$relativetime && $timezone !== $testNow->getTimezone()) {
+        $relativeTime = self::isTimeExpression($time);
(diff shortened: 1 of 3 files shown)


src/Chronos.php
```diff
@@ -268,18 +268,18 @@ protected function createNative(
...
-        $relativetime = self::isTimeExpression($time);
-        if (!$relativetime && $timezone !== $testNow->getTimezone()) {
+        $relativeTime = self::isTimeExpression($time);
+        if (!$relativeTime && $timezone !== $testNow->getTimezone()) {
             $testNow = $testNow->setTimezone($timezone ?? date_default_timezone_get());
         }
 
         if ($relative) {
-            $testNow = $testNow->modify($time);
+            $testNow = $testNow->modify($time ?? 'now');
```
othercorey, line 282: We can probably just return the clone if $time is not set, but can look into that later.

Issue #382, opened 2023-04-09
ComparisonTrait::isSameDay() / ComparisonTrait::isSameMonth()


andrii-pukhalevych:
https://github.com/cakephp/chronos/blob/b130b9ba4931b63b064bd571d288abecc0b14924/src/Traits/ComparisonTrait.php#L406-L409
Why in `ComparisonTrait::isSameDay()`, instead of comparing day/month/year, there is called `FormattingTrait::toDateString()` with `FormattingTrait::format()` inside, that looks much expensive operation?

Should it be following:
 `return $this->year === $dt->year && $this->month === $dt->month && $this->day === $dt->day;` ?

And should not we add one more method `ComparisonTrait::isSameMonth()`? I need such method in current project, but it does not present.
That will return `$this->year === $dt->year && $this->month === $dt->month`?  It will be possible to use it inside `ComparisonTrait::isSameDay()`.

markstory:
> Why in ComparisonTrait::isSameDay(), instead of comparing day/month/year, there is called FormattingTrait::toDateString() with FormattingTrait::format() inside, that looks much expensive operation?

I'm not sure we have sound reasoning as to why those methods are implemented the way they are. A large amount of the chronos project was inherited from Carbon. Changing the implementation around seems safe to me, and would likely be more efficient.

> And should not we add one more method ComparisonTrait::isSameMonth()? I need such method in current project, but it does not present.

The only rationale I would have to not add more methods is that the Chronos is already a very wide class/interface. I don't feel that is a strong argument though.  Any chance you'd be willing to make pull requests for these changes?

ndm2:
I have serious doubts that this would perform better. Currently it's 4 nested method calls, 2 x `toDateString()`, and 2 x `format()`. Using magic properties would result in 12 nested method calls, 6 x `__get()`, and 6 x `format()`.

PR #383, opened 2023-04-12
Add ComparisonTrait::isSameMonth()/ComparisonTrait::isSameYear()

andrii-pukhalevych:
PR for #382

### DIFF
--- a/src/ChronosInterface.php
+++ b/src/ChronosInterface.php
@@ -2,15 +2,15 @@
 declare(strict_types=1);
 
 /**
- * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
  *
  * Licensed under The MIT License
  * Redistributions of files must retain the above copyright notice.
  *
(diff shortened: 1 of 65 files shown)


src/ChronosInterface.php
```diff
@@ -534,6 +534,22 @@ public function isLeapYear(): bool;
      */
     public function isSameDay(ChronosInterface $dt): bool;
 
+    /**
+     * Checks if the passed in date is the same month as the instance current month.
+     *
+     * @param \Cake\Chronos\ChronosInterface $dt The instance to check against.
+     * @return bool
+     */
+    public function isSameMonth(ChronosInterface $dt): bool;
```
ADmad, line 543: Interfaces can only we changed in major releases, so this needs to target 3.x.

andrii-pukhalevych: Can it be added to 2.x without change of interface but only implementation?

ADmad: Theoretically it can be but IMO this isn't an important enough addition to deviate from the interface.

andrii-pukhalevych: Ok, but can you help with rebase? For me it gives conflicts in files. Or should create new PR?

ADmad: Creating a new PR would be easier :)

PR #386, opened 2023-04-16
Default ChronosTime to server time

othercorey:
This fixes several issues:

- ChronosTime() should default to server time not 00:00:00 with support for Chronos test now
- Add static helper ChronosTime::parse() to match DateTime
- Add static helpers to create specific times
- Changed the setters to roll over to parent time instead of rolling over on the individual field

Rollover should work the same as DateTimeImmutable::setTime() where 120 minutes sets the hours to 02. The individual setters also rollover to parent values now.

### DIFF
--- a/src/ChronosTime.php
+++ b/src/ChronosTime.php
@@ -18,6 +18,10 @@ use DateTimeImmutable;
 use DateTimeInterface;
 use InvalidArgumentException;
 
+/**
+ * @psalm-immutable
+ * @psalm-consistent-constructor
+ */
 class ChronosTime
 {
     /**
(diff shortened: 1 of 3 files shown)


src/ChronosTime.php
```diff
@@ -221,11 +266,14 @@ public function setHours(int $hours): static
      */
     public function setTime(int $hours = 0, int $minutes = 0, int $seconds = 0, int $microseconds = 0): static
     {
+        $ticks = $hours * self::TICKS_PER_HOUR +
+            $minutes * self::TICKS_PER_MINUTE +
+            $seconds * self::TICKS_PER_SECOND +
+            $microseconds * self::TICKS_PER_MICROSECOND;
+        $ticks = static::mod($ticks, self::TICKS_PER_DAY);
```
othercorey, line 277: With a signed 64-bit integer we can support up to 100 million days rollover days before failing so this solution should be stable.

src/ChronosTime.php
```diff
@@ -143,11 +188,11 @@ public function getSeconds(): int
      */
     public function setSeconds(int $seconds): static
     {
-        $prevSecondsTicks = $this->ticks % self::TICKS_PER_MINUTE - $this->ticks % self::TICKS_PER_SECOND;
-        $newSecondsTicks = static::mod($seconds, 60) * self::TICKS_PER_SECOND;
+        $baseTicks = $this->ticks - ($this->ticks % self::TICKS_PER_MINUTE - $this->ticks % self::TICKS_PER_SECOND);
```
othercorey, line 195: Same calculation as before but we subtract from the total ticks here and call it base instead of doing it below.

src/ChronosTime.php
```diff
@@ -143,11 +188,11 @@ public function getSeconds(): int
...
+        $newTicks = static::mod($baseTicks + $seconds * self::TICKS_PER_SECOND, self::TICKS_PER_DAY);
```
othercorey, line 196: This now adds all the rollover to ticks before performing the mod() on ticks per day.

othercorey:
~I don't know why phpstan complains about static in ChronosTime only.~

Moved the tags to correct docblock.

RikAlc23:
Hello, I would like to create a Java program for Windows 10 that has a button which, when clicked, starts a timer and that when it approaches minutes 10, 11, 12, and 13, opens a pop-up window with large text and an alarm sound. Can anyone help me with the code or point me in the right direction to get started?

src/ChronosTime.php
```diff
@@ -51,21 +55,37 @@ class ChronosTime
...
+            $this->ticks = static::parseTime($time);
         } else {
-            $this->ticks = static::parseTime(is_string($time) ? $time : $time->format('H:i:s.u'));
+            $time ??= Chronos::getTestNow() ?? Chronos::now();
+            $this->ticks = static::parseTime($time->format('H:i:s.u'));
         }
     }
 
+    /**
+     * Copies time from onther instance or from string in the format HH[:.]mm or HH[:.]mm[:.]ss.u
```
markstory, line 77:
```suggestion
     * Copies time from another instance or from string in the format HH[:.]mm or HH[:.]mm[:.]ss.u
```

markstory (approved):
Looks good to me.

Issue #385, opened 2023-04-14
False Statement in 2.x Docs about Timezone Fixation


ravage84:
Both, the documentation for the version 1.x & 2.x state:

>  The time and timezone for these objects is always fixed to 00:00:00 UTC [...]

https://book.cakephp.org/chronos/1/en/index.html#date-objects
https://book.cakephp.org/chronos/2/en/index.html#date-objects

While this seems to be true for Chronos 1.x:

- CakePHP 3.10.5
- CakePHP Chronos 1.3.0
- ``$config['App']['defaultTimezone'] = 'Europe/Zurich';``

```` bash
Psy Shell v0.11.14 (PHP 7.4.33 — cli) by Justin Hileman
New version is available at psysh.org/psysh (current: v0.11.14, latest: v0.11.15)
> use Cake\Chronos\Date;
> $today = Date::today();
= Cake\Chronos\Date @1681430400 {#2871
    hasFixedNow: false,
    date: 2023-04-14 00:00:00.0 UTC (+00:00),
  }

> $today->modify('+1 hours');
= Cake\Chronos\Date @1681430400 {#2871
    hasFixedNow: false,
    date: 2023-04-14 00:00:00.0 UTC (+00:00),
  }
````

=> ``2023-04-14 00:00:00.0 UTC (+00:00)``

It apparently isn't true (anymore) for Chronos 2.x:

- CakePHP 4.4.12
- CakePHP Chronos 2.3.2
- ``$config['App']['defaultTimezone'] = 'Europe/Zurich';``

```` bash
Psy Shell v0.11.12 (PHP 7.4.33 — cli) by Justin Hileman
New version is available at psysh.org/psysh (current: v0.11.12, latest: v0.11.15)
> use Cake\Chronos\Date;
> $today = Date::today();
= Cake\Chronos\Date @1681423200 {#2884
    hasFixedNow: false,
    date: 2023-04-14 00:00:00.0 Europe/Zurich (+02:00),
  }

> $today->modify('+1 hours');
= Cake\Chronos\Date @1681423200 {#2884
    hasFixedNow: false,
    date: 2023-04-14 00:00:00.0 Europe/Zurich (+02:00),
  }
````

=> ``2023-04-14 00:00:00.0 Europe/Zurich (+02:00))``

Is this a documenatation error or  is some detail missing which could explain the observed behavior?

[...]

PR #388, opened 2023-04-16
Update docs for Date + timezone

markstory:
In 2.x the timezone used for `Date` and `MutableDate` has changed from UTC to the server local time. Update the docs to reflect that.

Refs #385


othercorey:
This was debated, but with the change definitely should just update the docs.

PR #389, opened 2023-04-17
Add timezone to ChronosTime::now()

othercorey:
This aligns with Chronos::now().

### DIFF
--- a/src/ChronosTime.php
+++ b/src/ChronosTime.php
@@ -16,6 +16,7 @@ namespace Cake\Chronos;
 
 use DateTimeImmutable;
 use DateTimeInterface;
+use DateTimeZone;
 use InvalidArgumentException;
 
 /**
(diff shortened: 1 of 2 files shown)


src/ChronosTime.php
```diff
@@ -78,10 +87,10 @@ public function __construct(Chronos|ChronosTime|DateTimeInterface|string|null $t
      *
      * Defaults to server time.
      *
-     * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosTime|\DateTimeInterface|string|null $time Time
+     * @param \Cake\Chronos\Chronos|\Cake\Chronos\ChronosTime|\DateTimeInterface|string $time Time
      * @return static
      */
-    public static function parse(Chronos|ChronosTime|DateTimeInterface|string|null $time = null): static
+    public static function parse(Chronos|ChronosTime|DateTimeInterface|string $time): static
```
othercorey, line 93: Not sure if should support null and add timezone here as well.

PR #392, opened 2023-04-24
2.next: fix setISODate

LordSimal:
just tried the current `4.next` state of cakephp/cakephp (which is basically 4.5) and noticed, that my app broke due to this default value overwrite.

see https://www.php.net/manual/en/datetimeimmutable.setisodate.php


LordSimal:
CI will work after the 2.x => 2.next PR is merged and this branch is rebased.

PR #398, opened 2023-05-02
3.x: update stan

LordSimal:
That `private static function isRelativeOnly` was removed because it was not used anywhere. Is that correct @othercorey?


othercorey:
I don't remember why it wasn't needed. I'd have to check the changes again.

Issue #410, opened 2023-09-10
Reasoning around no longer extending native DateTimeImmutable


acelaya:
I was looking at chronos 3 changelog, and saw it no longer extends native date types.

I actually knew this was coming, as I have been using this project for quite some time, but I wanted to ask if it would be possible to share the reasoning of this architectural change.

Knowing this will help understand how maintainers envision this library to be used, and help migrating projects from v2 to v3. I can think on some good reasons, but I don't want to make any assumption.

Also, apologies if this is documented somewhere. I have been looking for issues, discussions and announcements and was not able to find this piece of information.

acelaya:
Actually, just found this, which explains it to some extent https://github.com/cakephp/chronos/blob/3cbb9e2d8af1065c7189e1dc3f4fa641e797719f/docs/en/2-4-upgrade-guide.rst#no-longer-extending-datetime

Antarian:
I understand it will make Chronos library cleaner and that is already decided... but:
- DateTimeImmutable did not required special treatment for Doctrine entity fields. I need to test this, but expect that Doctrine will now require custom converter for Chronos to DateTimeImmutable and back.
- declaring field DateTimeImutable or DateTimeInterface does not require any explanation. Declaring it Chronos does. ChronosDateTime would be slightly better.

Will stick to 2.* for now.

ADmad:
> DateTimeImmutable did not required special treatment for Doctrine entity fields. I need to test this, but expect that Doctrine will now require custom converter for Chronos to DateTimeImmutable and back.

Wouldn't this be solvable with a custom [mapping type class](https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/cookbook/advanced-field-value-conversion-using-custom-mapping-types.html)? CakePHP ORM uses similar type mapping classes and they are simple enough.

ADmad:
In light of the issues users are having upgrading to v3, we have tried reverting `Chronos` to extending `DatetimeImmutable` https://github.com/cakephp/chronos/pull/417. You can try out the [extend-datetimeimmutable](https://github.com/cakephp/chronos/tree/extend-datetimeimmutable) branch and share your feedback.

`ChronosDate` will remain unchanged and won't extend `DatetimeImmutable`.

othercorey:
I'll try to explain the reasons for the changes. First, thank you for opening this ticket. It's a very large change and very unexpected for most so you could have posted something much worse :).

[...]

PR #417, opened 2023-09-16
Change only Chronos to extend DateTimeImmutable once again

othercorey:
The original reasons for this change are described here: https://github.com/cakephp/chronos/issues/410#issuecomment-1722185717

- [x] Need to improve unit tests with more non-Chronos arguments now that the type declarations are relaxed.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -20,6 +20,7 @@ use DateTimeImmutable;
 use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
+use ReturnTypeWillChange;
 
 /**
  * An Immutable extension on the native DateTime object.
@@ -57,7 +58,7 @@ use InvalidArgumentException;
  * @psalm-immutable
  * @psalm-consistent-constructor
  */
-class Chronos
+class Chronos extends DateTimeImmutable
 {
     use FormattingTrait;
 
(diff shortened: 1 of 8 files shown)


LordSimal (approved):

ADmad (approved):

othercorey:
Removed unnecessary Chronos type declaration from formatter methods.

src/Chronos.php
```diff
@@ -998,7 +970,7 @@ public function format(string $format): string
      */
     public function getOffset(): int
     {
-        return $this->native->getOffset();
+        return parent::getOffset();
```
markstory, line 974: Is there much point in retaining methods like these if they only call parent?

markstory (commented):
Would we want to release this as 3.1.0?

src/Chronos.php
othercorey: Only if we want to ensure the type hints are available. I think most IDE's can figure this class out though.

I was thinking about commenting out the ones we don't need to wrap for types, but not delete them. We never know what PHP 9 will bring.

othercorey:
> Would we want to release this as 3.1.0?

Up to you. It would make it possible for people to transition without forcing it on them, but we do want this to be official. However, the only feedback we've seen was initially negative (or at least not positive) to not extending DateTimeImmutable.

CakePHP 5 deps need to be updated to only support this.

src/Chronos.php
```diff
@@ -2574,25 +2558,15 @@ public static function fromNow(Chronos|DateTimeInterface $other): DateInterval|b
...
 
-    /**
-     * Returns the time as a DateTimeImmutable instance.
-     *
-     * @return \DateTimeImmutable
-     */
-    public function toNative(): DateTimeImmutable
-    {
-        return $this->native;
-    }
```
shadowhand, line 2594: This method was **REMOVED** in a **BUGFIX** release?! This change would have broken production releases if our automated testing didn't catch it.

markstory: The removal was a mistake, we'll get that fixed.

PR #421, opened 2023-09-18
add halfOfYear, isFirstHalfOfYear, IsSecondHalfOfYear; improve getter…

brenoroosevelt:
I would like to propose the inclusion of the `halfOfYear` property and the `isFirstHalfOfYear` and `isSecondHalfOfYear` methods. The first half of the year corresponds to the first 6 months, while the second half corresponds to the remaining months.

Furthermore, I've made some improvements to the type hinting of the read-only properties in order to specify the numeric range. For example, changing `daysInMonth` from `int` to `int<1, 31>`. This allows for better static code analysis by PHPStan and Psalm.

Thank you in advance.

### DIFF
--- a/src/ChronosDate.php
+++ b/src/ChronosDate.php
@@ -30,15 +30,16 @@ use InvalidArgumentException;
  *
  * @property-read int $year
  * @property-read int $yearIso
- * @property-read int $month
- * @property-read int $day
- * @property-read int $dayOfWeek 1 (for Monday) through 7 (for Sunday)
- * @property-read int $dayOfYear 0 through 365
- * @property-read int $weekOfMonth 1 through 5
- * @property-read int $weekOfYear ISO-8601 week number of year, weeks starting on Monday
- * @property-read int $daysInMonth number of days in the given month
+ * @property-read int<1, 12> $month
+ * @property-read int<1, 31> $day
+ * @property-read int<1, 7> $dayOfWeek 1 (for Monday) through 7 (for Sunday)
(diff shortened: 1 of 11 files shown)


src/ChronosDate.php
```diff
@@ -1566,6 +1587,9 @@ public function __get(string $name): string|float|int|bool
             case $name === 'quarter':
                 return (int)ceil($this->month / 3);
 
+            case $name === 'halfOfYear':
+                return $this->month <= 6 ? 1 : 2;
```
othercorey, line 1591: Since we have `quarter`, maybe just `half` makes sense.

brenoroosevelt:
@othercorey  Changes made. renamed to 'half'

tests/TestCase/Date/GettersTest.php
```diff
@@ -0,0 +1,41 @@
...
+declare(strict_types=1);
+
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @copyright     Copyright (c) Brian Nesbitt <brian@nesbot.com>
```
markstory: We don't need to add this copyright tag to new files, only preserve it where it is.

othercorey:
Thanks!

PR #423, opened 2023-09-23
Align Chronos::getTimezone() return type with DateTimeImmutable

othercorey:
This avoids breaking compatibility and the need for an attribute.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -20,7 +20,7 @@ use DateTimeImmutable;
 use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
-use ReturnTypeWillChange;
+use RuntimeException;
 
 /**
  * An Immutable extension on the native DateTime object.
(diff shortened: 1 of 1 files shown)


src/Chronos.php
```diff
@@ -1009,12 +1009,16 @@ public function setTimezone(DateTimeZone|string $value): static
...
+     * @return \DateTimeZone
      */
-    #[ReturnTypeWillChange]
-    public function getTimezone(): ?DateTimeZone
+    public function getTimezone(): DateTimeZone
     {
-        return parent::getTimezone() ?: null;
+        $tz = parent::getTimezone();
+        if ($tz === false) {
+            throw new RuntimeException('Time zone could not be retrieved.');
```
othercorey, line 1018: I don't think it's possible to trigger this and just reflects the internal timelib functions. Maybe if you serialized DateTimeImmutable with a time zone that was later removed? I'm not sure.

markstory: Meh, I wouldn't worry about getting coverage on this case.

PR #424, opened 2023-09-24
Rename toNative() toDateTimeImmutable()

othercorey:
The term "native" made sense internally only. This is more explicit and easier to find.

### DIFF
--- a/src/ChronosDate.php
+++ b/src/ChronosDate.php
@@ -1564,11 +1564,11 @@ class ChronosDate
     }
 
     /**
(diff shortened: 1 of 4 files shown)


dereuromark:
Isnt that bc breaking? Native sounded fine to me.
Also, immutable should not be in there, leaking info

othercorey:
It technically is although it was just added. Can add toNative() as an alias to ensure no breaks.

othercorey:
I should have linked where this comes from - https://github.com/cakephp/chronos/issues/410#issuecomment-1726975219

PR #425, opened 2023-09-24
Support creating Chronos from ChronosTime

othercorey:


tests/TestCase/DateTime/ConstructTest.php
```diff
@@ -151,6 +152,24 @@ public function testCreateFromChronosDate()
         $this->assertSame('2021-01-01 00:00:00', $chronos->format('Y-m-d H:i:s'));
     }
 
+    public function testCreateFromChronosTime()
+    {
+        $time = new ChronosTime('20:14:12.123456');
+        $chronos = new Chronos($time);
+        $this->assertSame('20:14:12.123456', $chronos->format('H:i:s.u'));
```
markstory: Should we also have assertions for the date component of `Chronos` to help prevent regressions?

othercorey: It will always be the current date. I could check that?

othercorey: Should we special case using the test now date with a ChronosTime time or should it always be the current date?

markstory: I think having a test that it is the current date would be good for future us.

othercorey: Extended tests to compare full Chronos string.

markstory:
Looks good. I don't know what is up with psalm.

PR #426, opened 2023-09-24
Default Chronos::create() time components to 0

### LINKED ISSUE #419: RFC - Changing default time values for Chronos::create()
We would like to get feedback from the community on an idea for the time values in `Chronos::create()` for Chronos 3.

Current Behavior
---

The signature for `Chronos::create()` currently looks like this:
```php
create(?int $year = null, ?int $month = null, ?int $day = null, ?int $hour = null, ?int $minute = null, ?int $second = null, ?int $microsecond = null, DateTimeZone|string|null $timezone = null):
```

The current behavior is:

* When `$hour` is `null`, the current time is used for any time component that is also `null`. 
* When `$hour` is `0`,  0 is used for any time component that is also `null`.

Examples:

```php
Chronos::create(2023, 1, 1);
```

Will use `now` for the hour, minute second and microsecond.

```php
Chronos::create(2023, 1, 1, 0);
```

Will use `0` for the hour, minute second and microsecond.

Issues
---

* With named parameters, if you set only the `$minute` argument to 0, the hour, second and microsecond would still use `now`. This feels counter-intuitive.
  ```php
  Chronos::create(2023, 1, 1, minute: 0); // this will create something like "2023-01-01 13:00:02.125523"
  ```
* This behavior is different from constructing a `DateTimeImmutable` without including the time components:
  ```php
  new DateTimeImmutable('2023-01-01'); // this will create "2023-01-01 00:00:00"
  ```

Proposal
---

Change the default values for `$hour`, `$minute`, `$second` and `$microsecond` to `0`. 

Users can still specify `null` to get `now`, but they will have to specify it for all time components. There are simpler ways to create a Chronos instance of `now` so this doesn't feel like a harsh requirement.

othercorey:
closes https://github.com/cakephp/chronos/issues/419

This expands the scope of the RFC a little since other wrappers that call Chronos::create() were also defaulting to `null`.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -563,10 +563,10 @@ class Chronos extends DateTimeImmutable
         ?int $year = null,
         ?int $month = null,
         ?int $day = null,
-        ?int $hour = null,
-        ?int $minute = null,
-        ?int $second = null,
-        ?int $microsecond = null,
+        ?int $hour = 0,
+        ?int $minute = 0,
+        ?int $second = 0,
+        ?int $microsecond = 0,
         DateTimeZone|string|null $timezone = null
     ): static {
         $now = static::now();
@@ -574,16 +574,10 @@ class Chronos extends DateTimeImmutable
         $month = $month ?? $now->format('m');
         $day = $day ?? $now->format('d');
 
-        if ($hour === null) {
-            $hour = $now->format('H');
-            $minute = $minute ?? $now->format('i');
-            $second = $second ?? $now->format('s');
-            $microsecond = $microsecond ?? $now->format('u');
-        } else {
-            $minute = $minute ?? 0;
-            $second = $second ?? 0;
-            $microsecond = $microsecond ?? 0;
-        }
+        $hour = $hour ?? $now->format('H');
+        $minute = $minute ?? $now->format('i');
+        $second = $second ?? $now->format('s');
+        $microsecond = $microsecond ?? $now->format('u');
 
         $instance = static::createFromFormat(
             'Y-m-d H:i:s.u',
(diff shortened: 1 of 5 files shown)


othercorey:
The diff() test changes are due to a limitation in DatePeriod() where the end date isn't included when it's on exactly 00:00:00.

There is a new option in PHP 8.2 to include the end date that we should probably expose in the diff wrappers.

src/Chronos.php
```diff
@@ -563,27 +563,21 @@ public static function create(
         ?int $year = null,
         ?int $month = null,
         ?int $day = null,
-        ?int $hour = null,
-        ?int $minute = null,
-        ?int $second = null,
-        ?int $microsecond = null,
+        ?int $hour = 0,
```
ADmad, line 566: So time parts now default to `0` but the date parts still default to `null` aka based on current date? Doesn't that seem inconsistent?

othercorey: The date seems like it should be required in both create() and createFromDate(), yes?

markstory: Having the date parts default to 0 would be pretty unpleasant and a larger break in compatibility. 

> The date seems like it should be required in both create() and createFromDate()

Defaulting to 'now' is a good default value that we should continue to use.

src/Chronos.php
```diff
@@ -563,27 +563,21 @@ public static function create(
...
-            $microsecond = $microsecond ?? $now->format('u');
-        } else {
-            $minute = $minute ?? 0;
-            $second = $second ?? 0;
-            $microsecond = $microsecond ?? 0;
-        }
+        $hour = $hour ?? $now->format('H');
+        $minute = $minute ?? $now->format('i');
+        $second = $second ?? $now->format('s');
+        $microsecond = $microsecond ?? $now->format('u');
```
markstory, line 580: Having all the time components default to `0` will be a bigger break than defaulting all values to 'now'. Would we be in a better place if we removed the special scenario for `hour: 0`? Having each component of the time default to the present, would make using named parameters preserve as much compatibility with what we have today, and increase consistency between how the date & time component parameters work.

othercorey: The reason for this change is to allow calls like `Chronos::create(2023, 01, 01, minute: 10)` Currently that will create an instance where hour, second and microsecond are now and minute is 10.

PR #427, opened 2023-09-24
Bring back toNative() alias of toDateTimeImmutable()

othercorey:
refs https://github.com/cakephp/chronos/pull/424

We want to keep BC with 3.0 as users have changed code to call `toNative()`. Users can decide which makes their code clearer or cleaner.

### DIFF
--- a/src/ChronosDate.php
+++ b/src/ChronosDate.php
@@ -1574,6 +1574,18 @@ class ChronosDate
     }
 
     /**
+     * Returns an `DateTimeImmutable` instance set to this clock time.
+     *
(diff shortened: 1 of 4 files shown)


dereuromark:
Ohh Chronos itself is missing here it seems

othercorey:
Chrnis is already DateTimeImmutable

othercorey:
But it could be a BC break although its a weird conversation

dereuromark:
But it still isnt the actual native object, so it does make sense to keep toNative() everywhere if someone really needs the native class type.
Especially for BC reasons, though, this method for sure can only be deprecated now, not removed.

PR #428, opened 2023-09-24
Fix casting to string for ChronosDate and ChronosTime.

ADmad:
Make all classes explicitly implemement Stringable as recommeneded by the PHP docs.

### DIFF
--- a/src/ChronosTime.php
+++ b/src/ChronosTime.php
@@ -18,12 +18,13 @@ use DateTimeImmutable;
 use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
+use Stringable;
 
 /**
  * @psalm-immutable
  * @psalm-consistent-constructor
  */
-class ChronosTime
+class ChronosTime implements Stringable
 {
     /**
      * @var int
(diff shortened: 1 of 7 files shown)


src/FormattingTrait.php
```diff
@@ -32,7 +32,7 @@ trait FormattingTrait
      */
     public static function resetToStringFormat(): void
     {
-        static::setToStringFormat(Chronos::DEFAULT_TO_STRING_FORMAT);
+        static::setToStringFormat(static::DEFAULT_TO_STRING_FORMAT);
```
ADmad, line 35: Before this fix and addition of the constant to `ChronosDate`, calling `ChronosDate::__toString()` after `ChronosDate::resetToStringFormat()` returned value in `Y-m-d H:i:s` format instead of `H:i:s`.

src/ChronosTime.php
```diff
@@ -323,6 +337,37 @@ public function format(string $format): string
         return $this->toDateTimeImmutable()->format($format);
     }
 
+    /**
+     * Reset the format used to the default when converting to a string
+     *
+     * @return void
+     */
+    public static function resetToStringFormat(): void
```
othercorey, line 345: We probably never needed the reset helpers since you can pass the constant to set directly.

ADmad: I just made it consistent with the other 2 classes.

PR #429, opened 2023-09-24
Allow passing DatePeriod options to diffFiltered()

othercorey:
This allows users to pass the new PHP 8.2 option that includes the end date in the period. This is helpful when iterating over dates that end on midnight as that day won't be included.


markstory (approved):
Seems good to me.

PR #430, opened 2023-09-25
Allow setting time zone when converting to DateTimeImmutable

othercorey:
Allow users to set the time zone when converting to DateTimeImmutable instead of only defaulting to the internal implementation.

Users will probably use this when converting between systems and will need flexibility.

This adds no extra overhead to the conversion.

### DIFF
--- a/tests/TestCase/Date/StringsTest.php
+++ b/tests/TestCase/Date/StringsTest.php
@@ -176,9 +176,15 @@ class StringsTest extends TestCase
     public function testToDateTimeImmutable(): void
     {
         $d = ChronosDate::now();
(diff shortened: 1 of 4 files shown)


tests/TestCase/Date/StringsTest.php
```diff
@@ -176,9 +176,15 @@ public function testToW3cString()
     public function testToDateTimeImmutable(): void
     {
         $d = ChronosDate::now();
-        $this->assertSame($d->format('Y-m-d'), $d->toDateTimeImmutable()->format('Y-m-d'));
+        $this->assertSame($d->format('Y-m-d H:i:s.u'), $d->toDateTimeImmutable()->format('Y-m-d H:i:s.u'));
```
othercorey, line 179: Extend the test to ensure the time isn't warped.

PR #433, opened 2023-10-02
Restore Chronos::toNative()

### LINKED ISSUE #432: Method toNative() has gone missing
https://github.com/cakephp/chronos/pull/417#pullrequestreview-1652711214

According to Semantic Versioning, methods and attributes MAY NOT be removed except in major releases. The `toNative()` method was present in versions 3.0.0 and 3.0.1, but removed in 3.0.2. This breaks expectations and trust in versioning.

markstory:
This was accidently removed in the shuffle of #417 and #427

Fixes #432

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -2581,6 +2581,18 @@ class Chronos extends DateTimeImmutable implements Stringable
     }
 
     /**
+     * Returns a DateTimeImmutable instance
+     *
+     * This method returns a PHP DateTimeImmutable without Chronos extensions.
+     *
+     * @return \DateTimeImmutable
+     */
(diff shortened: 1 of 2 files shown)


dereuromark (approved):

othercorey:
I wonder if we should deprecate this now that Chronos is a DateTimeImmutable.

dereuromark:
To make it more consistent, maybe we could deprecated it same way? To immutableDatetime() conversions method?
Or just deprecate it.

othercorey:
If it works for people, we can just leave it. No need to force them to deal with the mess. Maybe it will naturally phase out.

PR #439, opened 2023-10-14
Add ChronosTime::endOfDay()

othercorey:


PowerKiKi:
Adding a feature like this should have been released as a minor version, not patch version. Otherwise, it makes it harder for consumers to declare dependencies on specific _features_. Same goes for #417.

PR #443, opened 2023-10-24
Provide PSR-20 Clock implementation

odan:
This PR provides a [PSR-20](https://www.php-fig.org/psr/psr-20/) compliant Clock implementation for the Chronos package.

The PSR-20 Clock interface defines a standard way to interact with system time, allowing for better testability and flexibility.

### DIFF
--- a/src/ClockFactory.php
+++ b/src/ClockFactory.php
@@ -22,7 +22,7 @@ use Psr\Clock\ClockInterface;
 /**
  * PSR-20 Clock implementation.
  */
-class ChronosClock implements ClockInterface
+class ClockFactory implements ClockInterface
 {
     private DateTimeZone|string|null $timezone;
 
--- a/src/ClockFactory.php
+++ b/src/ClockFactory.php
@@ -10,7 +10,6 @@ namespace Cake\Chronos;
  * Redistributions of files must retain the above copyright notice.
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- * @copyright     Copyright (c) Daniel Opitz
  * @link          https://cakephp.org CakePHP(tm) Project
  * @license       https://www.opensource.org/licenses/mit-license.php MIT License
  */
--- a/composer.json
+++ b/composer.json
@@ -25,11 +25,15 @@
         "source": "https://github.com/cakephp/chronos"
     },
     "require": {
-        "php": ">=8.1"
+        "php": ">=8.1",
+        "psr/clock": "^1.0"
     },
     "require-dev": {
         "cakephp/cakephp-codesniffer": "^5.0",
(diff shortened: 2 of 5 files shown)


othercorey:
I am away right now but I think the term ChronosClock is misleading and confusing compared to the other classes. Can we expose this some other way?

odan:
The idea was to name it `ChronosClock` because it returns a Chronos instance, that is also a TimeTimeImmutable instance.

Other class names: `Clock`, `ClockFactory` or `DateTimeImmutableFactory`.

odan:
After thinking about it, I guess my approach doesn't make sense from a Chronos perspective, because if you are already using Chronos, you may not need the ClockInterface at all.

markstory:
> After thinking about it, I guess my approach doesn't make sense from a Chronos perspective, because if you are already using Chronos, you may not need the ClockInterface at all.

I think it is another way for us to participate in the community standards within the larger PHP ecosystem. That alone has value I think.

othercorey:
What  if we implemented it with an anonymous class that's returned by a static helper. That would provide easy access to the interface, right?

odan:
> Error: Type DateTimeImmutable is not used in this file.

phpcs complains about this problem even though Chronos is inherited from DateTimeImmutable.

> What if we implemented it with an anonymous class that's returned by a static helper. That would provide easy access to the interface, right?

A static method cannot then be used like an interface, for example when using dependency injection.

othercorey:
ChronosFactory or ClockFactory should work then.

markstory:
There is one last lint error to fix and then we should be able to move forward with this.

othercorey:
The class needs to be renamed to ClockFactory.

src/ClockFactory.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Cake\Chronos;
+
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @copyright     Copyright (c) Daniel Opitz
```
othercorey: We don't need  separate copyright, just the cakephp one.

odan: I added this, because it was also added here:

https://github.com/cakephp/chronos/blob/3.x/src/Chronos.php#L11
https://github.com/cakephp/chronos/blob/3.x/src/DifferenceFormatter.php#L11

Maybe we remove the other redundant copyright tags as well?

othercorey: The very original Chronos was forked from a different project so Mark inherited a copyright for some of the files. It's very close to not applying now, but definitely not new code.

PR #446, opened 2023-10-29
Fix off by one errors in diffFiltered

markstory:
Remove the work around that I added and add tests covering `DifferenceTrait` in `Date` classes. A similar fix will likely be needed in 3.x

### DIFF
--- a/src/Traits/DifferenceTrait.php
+++ b/src/Traits/DifferenceTrait.php
@@ -191,7 +191,7 @@ trait DifferenceTrait
 
         $period = new DatePeriod($start, $ci, $end);
         $vals = array_filter(iterator_to_array($period), function (DateTimeInterface $date) use ($callback) {
-            return $callback(static::instance($date));
+            return $callback(static::parse($date));
         });
 
         $diff = count($vals);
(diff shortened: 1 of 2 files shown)


othercorey:
I'm not sure this fixes it. It's more like moving the plug from one leak to another. In 3.x, I tried to address it by allowing users to pass DatePeriod options into to the filtered wrappers.

The issue will still occur if a specific date range happens. If we were on PHP 8.2, I would have pushed to use https://www.php.net/manual/en/class.dateperiod.php#dateperiod.constants.include-end-date by default.

othercorey:
See https://github.com/cakephp/chronos/pull/429

markstory:
> I'm not sure this fixes it. It's more like moving the plug from one leak to another

I don't disagree, but this change should fix the regression which is what I think 2.x users are looking for.

> If we were on PHP 8.2, I would have pushed to use include-end-date

I agree this is the better solution as we can let users decide where their end date should be.

othercorey:
Do you want to backport the 3.x changes with this?

markstory:
> Do you want to backport the 3.x changes with this?

I can take care of that.

markstory:
> Do you want to backport the 3.x changes with this?

@othercorey I had time to look at this today and I don't think that we'll be able to backport the additional option parameter as it would require changing `ChronosInterface::diffFiltered` which is a breaking change :cry:

othercorey:
Ah ok. I had hoped it was easy, but users can just move to 3.x.

PR #447, opened 2023-11-01
Improve exception messages

dereuromark:
```
1) Data\Test\TestCase\Model\Table\AddressesTableTest::testTouch
InvalidArgumentException: Time string is not in expected format: "HH[:.]mm" or "HH[:.]mm[:.]ss.u".

/home/runner/work/cakephp-data/cakephp-data/vendor/cakephp/chronos/src/ChronosTime.php:122
```

hard to understand

it is possible to give the values for quicker debugging however.

```
InvalidArgumentException: Time string `-2 seconds` is not in expected format `HH[:.]mm` or `HH[:.]mm[:.]ss.u`.
```

fixed also some formatting around code fencing.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -936,7 +936,7 @@ class Chronos extends DateTimeImmutable implements Stringable
     {
         $new = parent::modify($modifier);
         if ($new === false) {
-            throw new InvalidArgumentException('Unable to modify date using: ' . $modifier);
+            throw new InvalidArgumentException(sprintf('Unable to modify date using `%s', $modifier));
         }
 
         return $new;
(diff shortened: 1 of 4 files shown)


src/Chronos.php
```diff
@@ -936,7 +936,7 @@ public function modify(string $modifier): static
     {
         $new = parent::modify($modifier);
         if ($new === false) {
-            throw new InvalidArgumentException('Unable to modify date using: ' . $modifier);
+            throw new InvalidArgumentException(sprintf('Unable to modify date using `%s', $modifier));
```
garas:
```suggestion
            throw new InvalidArgumentException(sprintf('Unable to modify date using `%s`', $modifier));
```

PR #454, opened 2024-07-13
Fix error on PHP 8.4

ADmad:
PHP 8.4 adds a new interface method DatetimeImmutable::createFromTimestamp().

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -736,13 +736,23 @@ class Chronos extends DateTimeImmutable implements Stringable
     /**
      * Create an instance from a timestamp
      *
-     * @param int $timestamp The timestamp to create an instance from.
+     * @param float|int $timestamp The timestamp to create an instance from.
      * @param \DateTimeZone|string|null $timezone The DateTimeZone object or timezone name the new instance should use.
      * @return static
      */
-    public static function createFromTimestamp(int $timestamp, DateTimeZone|string|null $timezone = null): static
+    public static function createFromTimestamp(float|int $timestamp, DateTimeZone|string|null $timezone = null): static
     {
-        return static::now($timezone)->setTimestamp($timestamp);
+        if (PHP_VERSION_ID >= 80400 && $timezone === null) {
+            return parent::createFromTimestamp($timestamp);
+        }
+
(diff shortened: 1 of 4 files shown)


ADmad:
The signature mismatch issue on PHP 8.4 is easily fixable but as seen from the failed tests the result of `DateTimeImmutable::createFromTimestamp()` is different than our implementation.

ADmad:
The change in behavior will need to be documented in the release notes. If `$timezone` is not explicitly passed then the instance has timezone set to `+00:00` unlike earlier where the currently set default timezone was used.

src/Chronos.php
```diff
@@ -736,13 +736,15 @@ public static function createFromArray(array $values): static
...
-     * @param int $timestamp The timestamp to create an instance from.
+     * @param float|int $timestamp The timestamp to create an instance from.
      * @param \DateTimeZone|string|null $timezone The DateTimeZone object or timezone name the new instance should use.
      * @return static
      */
-    public static function createFromTimestamp(int $timestamp, DateTimeZone|string|null $timezone = null): static
+    public static function createFromTimestamp(float|int $timestamp, DateTimeZone|string|null $timezone = null): static
     {
-        return static::now($timezone)->setTimestamp($timestamp);
+        $instance = PHP_VERSION_ID >= 80400 ? parent::createFromTimestamp($timestamp) : new static('@' . $timestamp);
```
othercorey, line 745: Now that we accept float, do we need to do any special rounding/etc to mimic the 8.4 behavior when converting to a @timestamp ?

I hope not. I assume they expect the timestamp to still be an integer in float format.

ADmad: > I assume they expect the timestamp to still be an integer in float format.

No, the fractional part of floats is used as microseconds https://3v4l.org/OT84B#v8.1.0
We don't have to do any additional processing.

othercorey: ah, ok. Wasn't sure if the @ format would throw an error, but I think that's ok if it does.

othercorey (approved):

markstory:
I started a draft release for 3.1 with the behavior change we've inherited from PHP.

PR #458, opened 2024-07-22
Added replacement suggestions for some deprecated functions

Staubiii:
We noticed some non conclusive Deprecation Warning messages for certain functions and added replacement suggestions where applicable.


markstory:
Thank you 🎉

PR #459, opened 2024-07-24
Fix deprecations from internals for Date

### LINKED ISSUE #453: Invalid deprecation error for Date::startOfWeek()
I just updated my installation of Chronos to the latest available version for the project (PHP 7.3, so best I get is 2.4.4), and I got this nonsensical error:
`(new Date($anyDateHere))->addWeeks(1)->startOfWeek()`
throws
`2.5 startOfDay() will be removed in 3.x.`
But this method calls `startOfDay()` implicitly, under the hood, users can't NOT call it.

markstory:
Fix deprecations for startOfDay/endOfDay being called by startOfWeek/endOfWeek. Because dates don't have times we don't need to modify any time related components and can just return early.

Fixes #453

### DIFF
--- a/src/Traits/ModifierTrait.php
+++ b/src/Traits/ModifierTrait.php
@@ -1051,6 +1051,9 @@ trait ModifierTrait
         if ($dt->dayOfWeek !== static::$weekStartsAt) {
             $dt = $dt->previous(static::$weekStartsAt);
         }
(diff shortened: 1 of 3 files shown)


dereuromark:
Phpstan seems to fail but unrelated

PR #460, opened 2024-07-26
Fix wrong return types

Harfusha:
Fixed return types of "date finding methods" that had return type mixed


dereuromark (approved):

markstory:
Thank you 🎉

PR #467, opened 2024-11-18
Fix fatal error on PHP 8.4

gersonfs:
I'm trying to use cakephp 4.x with php 8.4 and but this signature is causing problems. 
https://php.watch/versions/8.4/datetime-createFromTimestamp
`Fatal error: Declaration of Cake\Chronos\Traits\FactoryTrait::createFromTimestamp(int $timestamp, $tz = null): Cake\Chronos\ChronosInterface must be compatible with DateTimeImmutable::createFromTimestamp(int|float $timestamp): static`

### DIFF
--- a/src/Traits/FactoryTrait.php
+++ b/src/Traits/FactoryTrait.php
@@ -336,7 +336,8 @@ trait FactoryTrait
      * @param \DateTimeZone|string|null $tz The DateTimeZone object or timezone name the new instance should use.
      * @return static
      */
-    public static function createFromTimestamp(int $timestamp, $tz = null): ChronosInterface
+    #[ReturnTypeWillChange]
+    public static function createFromTimestamp(int|float $timestamp, $tz = null): ChronosInterface
     {
         $instance = static::now($tz)->setTimestamp($timestamp);
         if (get_class($instance) === ChronosDate::class) {


src/Traits/FactoryTrait.php
```diff
@@ -336,7 +336,8 @@ public static function createFromArray(array $values): ChronosInterface
      * @param \DateTimeZone|string|null $tz The DateTimeZone object or timezone name the new instance should use.
      * @return static
      */
-    public static function createFromTimestamp(int $timestamp, $tz = null): ChronosInterface
+    #[ReturnTypeWillChange]
+    public static function createFromTimestamp(int|float $timestamp, $tz = null): ChronosInterface
```
gersonfs, line 340: Oops, we have tests for php 7.4, any suggestions on how to maintain compatibility?

LordSimal:
This was only fixed in Chronos v3 as you can see here:
https://github.com/cakephp/chronos/pull/454

I'd recommend upgrading to CakePHP 5 before trying to get CakePHP 4 running with PHP 8.4

With that you will get [Chronos v3](https://github.com/cakephp/cakephp/blob/5.0.0/composer.json#L29) and therefore the fix for PHP 8.4

gersonfs:
Ok, I'll have more work but that's it. Thanks

PR #472, opened 2025-06-14
Add ChronosPeriod which returns Chronos instances from DatePeriod

othercorey:
This is really just an iterator converter, but I don't think anything else is needed.

### DIFF
--- a/src/ChronosPeriod.php
+++ b/src/ChronosPeriod.php
@@ -1,0 +1,83 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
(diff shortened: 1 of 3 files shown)


markstory (approved):
Seems reasonable to me.

src/ChronosPeriod.php
```diff
@@ -0,0 +1,83 @@
...
+    protected Iterator $iterator;
+
+    /**
+     * @param \DatePeriod $period
+     */
+    public function __construct(DatePeriod $period)
+    {
+        /** @var \Iterator<int, \DateTimeInterface> $iterator */
+        $iterator = $period->getIterator();
+        $this->iterator = $iterator;
```
othercorey, line 41: We assume getIterator() actually returns an InternalIterator even though IteratorAggregate is typed as Traversable.

ADmad (approved):

Issue #471, opened 2025-05-19
Trait declaration mismatch on 8.4


adominodsq:
Running on PHP 8.4.5 gives the following error with chronos 2.4.5:

In FactoryTrait.php line 339:
                                                                                                                                                                                                                                
  Declaration of Cake\Chronos\Traits\FactoryTrait::createFromTimestamp(int $timestamp, $tz = null): Cake\Chronos\ChronosInterface must be compatible with DateTimeImmutable::createFromTimestamp(int|float $timestamp): static  



Updating FactoryTrait.php:339 to allow the timestamp to be an int or float fixes the issue.

dereuromark:
Are you using an outdated version of the library?
https://github.com/cakephp/chronos/blob/e5377b2797851efd3ffbd51a69528597bc166950/src/Chronos.php#L743

This was all documented and released with https://github.com/cakephp/chronos/releases/tag/3.1.0

dereuromark:
You could make a backport PR for 2.x (and 2.5.0) and we could discuss a fitting backport release.
It is technically not fully BC, so if someone already extended it they would run into issues.
So that long ago after the v2 releases I am not sure if that's a viable approach.

Is there a reason you cannot use v3?

In the meantime I updated the https://github.com/cakephp/chronos/wiki#version-map to reflect the current min/max.

dereuromark:
PR open: https://github.com/cakephp/chronos/pull/474

PR #474, opened 2025-06-18
Allow float timestamp for createFromTimestamp() to match php

othercorey:
fixes https://github.com/cakephp/chronos/issues/471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.4', '8.0', '8.1', '8.2']
+        php-version: ['8.4']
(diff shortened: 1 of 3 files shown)


ADmad (approved):

.github/workflows/ci.yml
```diff
@@ -17,11 +17,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.4', '8.0', '8.1', '8.2']
+        php-version: ['8.4']
```
ADmad, line 20: why are we only testing on 8.4 when the php constraint is `>=8.0`?

dereuromark: that was resolved using https://github.com/cakephp/chronos/pull/475/files

PR #476, opened 2025-06-19
Add ChronosDatePeriod

othercorey:
refs https://github.com/cakephp/chronos/pull/472

This allows any DatePeriod and converts to a ChronosDate instance.


othercorey:
@markstory Any concerns with this approach for ChronosDate?

PR #477, opened 2025-06-22
update stan + cleanup

LordSimal:
I don't know why the testsuite wasn't using our central template but I didn't find any discrepencies with what was present right now.


LordSimal:
That stan issue is correct imho.

This if can't be reached, since that property is only typed to be `DateTimeImmutable` and not `DateTimeImmutable|bool`.

.github/workflows/ci.yml
```diff
@@ -15,56 +15,8 @@ permissions:
...
-        if [[ ${{ matrix.php-version }} == '8.1' && ${{ matrix.dependencies }} == 'highest' ]]; then
-          export CODECOVERAGE=1 && vendor/bin/phpunit --display-deprecations --display-warnings --display-incomplete --display-skipped --coverage-clover=coverage.xml
-        else
-          vendor/bin/phpunit --display-deprecations --display-warnings
-        fi
-
-    - name: Code Coverage Report
-      if: success() && matrix.php-version == '8.1' && matrix.dependencies == 'highest'
-      uses: codecov/codecov-action@v5
+    uses: cakephp/.github/.github/workflows/testsuite-with-db.yml@5.x
```
othercorey: I don't think we need db

LordSimal: right

PR #481, opened 2025-08-18
Style: Removed redundant phpdoc var/return tags

arshidkv12:


dereuromark:
Even if not needed, it also doesn't hurt to keep them IMO
For some cases, e.g. array, it can also be counterproductive if we ever want to better annotate the actual iterable type on it (`array<...>`)

arshidkv12:
Thank you

PR #484, opened 2025-10-26
Add toRfc7231String() helper

othercorey:
This replaces the deprecated `DATE_RFC7231` constant. Added to Chronos only since it doesn't make much sense for ChronosDate even though we include all of FormattingTrait for backwards compatibility.

### DIFF
--- a/tests/TestCase/DateTime/StringsTest.php
+++ b/tests/TestCase/DateTime/StringsTest.php
@@ -160,6 +160,12 @@ class StringsTest extends TestCase
         $this->assertSame('1639224001', $time->toUnixString());
     }
 
+    public function testToRfc7231String()
+    {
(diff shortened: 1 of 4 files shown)


dereuromark:
>  No matching release found for phpstan!

I feel like we should get rid of phive. require-dev is just fine.

tests/TestCase/DateTime/StringsTest.php
```diff
@@ -160,6 +160,12 @@ public function testToUnixString()
         $this->assertSame('1639224001', $time->toUnixString());
     }
 
+    public function testToRfc7231String()
+    {
+        $time = Chronos::parse('2014-04-20 08:00:00');
```
ADmad: Could add a non UTC timezone to the string to show the method properly converts the time before formatting.

PR #487, opened 2025-10-29
Use match() instead of switch/case.

ADmad:
Also improve docblock type

### DIFF
--- a/src/FormattingTrait.php
+++ b/src/FormattingTrait.php
@@ -250,7 +250,7 @@ trait FormattingTrait
 
         trigger_error(
             'Using toQuarter() with `$range=true` is deprecated. Use `toQuarterRange()` instead.',
-            E_USER_DEPRECATED
+            E_USER_DEPRECATED,
         );
 
         return $this->toQuarterRange();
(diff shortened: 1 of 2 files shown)


dereuromark (approved):

src/FormattingTrait.php
```diff
@@ -259,23 +259,19 @@ public function toQuarter(bool $range = false): int|array
...
-            case 3:
-                return [$year . '-07-01', $year . '-09-30'];
-            default:
-                return [$year . '-10-01', $year . '-12-31'];
-        }
+        return match ($quarter) {
+            1 => [$year . '-01-01', $year . '-03-31'],
+            2 => [$year . '-04-01', $year . '-06-30'],
+            3 => [$year . '-07-01', $year . '-09-30'],
+            default => [$year . '-10-01', $year . '-12-31'],
```
garas:
```suggestion
            4 => [$year . '-10-01', $year . '-12-31'],
```
`$this->format('m')` is always '01' to '12'.
`$quarter` is always 1 to 4.

dereuromark: I wonder if phpstan requires a default catch all?
If so, maybe it should throw a invalid argument exception.

ADmad: phpstan often complains about missing `default` and the existing code was also using default case instead of `4`, hence copilot used the same. I'll see if using `4` keeps phpstan happy.

garas: Do annotating `$quarter` as `int<1, 4>` makes PHPStan happy? 
Adding impossible `default` just to make PHPStan happy is not what should be done when finite list is known.

ADmad: Done

PR #494, opened 2026-01-28
Fix createFromFormat to respect setTestNow for missing components

### LINKED ISSUE #452: Chronos::createFromFormat using a date without time does not respect Chronos::setTestNow
It seems [Chronos::setTestNow()](https://book.cakephp.org/chronos/3/en/index.html#testing-aids) only fixates the time and not the rest of a date, is that intended?

For example, when using Chronos::createFromFormat() with a partial format, things like the day and month are not fixated:

```php
Chronos::setTestNow(new Chronos('2020-12-01 00:00:00'));
$date = Chronos::createFromFormat('m-d', '10-5'); // I would expect this to become "2020-10-05 00:00:00
$date = Chronos::createFromFormat('d', '5'); // I would expect this to become "2020-12-05 00:00:00
```

Note this is similar to https://github.com/briannesbitt/Carbon/issues/1280

dereuromark:
## Summary

When `setTestNow()` is used for testing, `createFromFormat()` now uses values from the test time for any date/time components not present in the format string.

Previously, partial formats like `createFromFormat('m-d', '10-05')` would use the real system time for the year, making tests unreliable. Now it uses the testNow year.

The fix:
- Parses the format string to detect which components (year, month, day, hour, minute, second, microsecond) are present
- For missing components, uses values from testNow instead of real current time
- Properly handles escaped characters in format strings
- Respects `!` and `|` modifiers which intentionally reset unspecified components to Unix epoch/zero

### Example

```php
Chronos::setTestNow(new Chronos('2020-12-01 00:00:00'));
$date = Chronos::createFromFormat('m-d', '10-05');
// Before: year from real system time (unpredictable in tests)
// After: 2020-10-05 00:00:00 (year from testNow)
```

Closes #452

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -660,7 +660,94 @@ class Chronos extends DateTimeImmutable implements Stringable
             throw new InvalidArgumentException($message);
         }
 
+        $testNow = static::getTestNow();
+        if ($testNow !== null) {
+            $dateTime = static::applyTestNowToMissingComponents($dateTime, $format, $testNow);
+        }
+
         return $dateTime;
+    }
+
+    /**
+     * Apply testNow values to date/time components that weren't in the format string.
+     *
+     * @param static $dateTime The parsed datetime instance.
+     * @param string $format The format string used for parsing.
+     * @param \Cake\Chronos\Chronos $testNow The test now instance.
+     * @return static
+     */
+    protected static function applyTestNowToMissingComponents(
+        self $dateTime,
+        string $format,
+        Chronos $testNow,
+    ): static {
+        // Parse format string to find which characters are actual format specifiers (not escaped)
+        $formatChars = static::getFormatCharacters($format);
+
+        // Check which components are present in the format
+        $hasYear = (bool)array_intersect($formatChars, ['Y', 'y', 'o', 'X', 'x']);
+        $hasMonth = (bool)array_intersect($formatChars, ['m', 'n', 'M', 'F']);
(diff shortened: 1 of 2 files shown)


src/Chronos.php
```diff
@@ -660,9 +660,96 @@ public static function createFromFormat(
...
+        $hasMinute = (bool)array_intersect($formatChars, ['i']);
+        $hasSecond = (bool)array_intersect($formatChars, ['s']);
+        $hasMicro = (bool)array_intersect($formatChars, ['u', 'v']);
+
+        // If the format includes '!' or '|', PHP resets unspecified components to Unix epoch or zero
+        // In that case, we should not override with testNow
+        $hasReset = in_array('!', $formatChars, true) || in_array('|', $formatChars, true);
+        if ($hasReset) {
+            return $dateTime;
+        }
```
markstory, line 701: Why are we adding new syntax to PHP's datetime formats?

dereuromark: We're not adding new syntax - `!` and `|` are existing PHP DateTime format modifiers (see [PHP docs](https://www.php.net/manual/en/datetimeimmutable.createfromformat.php)).

The code is detecting their presence so we know to skip applying testNow values. When a user explicitly uses `!` or `|`, they're telling PHP to reset unspecified components to Unix epoch, so we should respect that intent rather than substituting testNow values.

ADmad: LOL.. never knew such modifiers existed.

dereuromark: Open for alternative solutions to solving this issue, though :)

markstory: > LOL.. never knew such modifiers existed.

Same. Carry on then.

Issue #444, opened 2023-10-26
ChronosInterval


dereuromark:
> trigger_error(
            'Since 2.4 ChronosInterval is deprecated. Use `Chronos::createInterval() instead.`',
            E_USER_DEPRECATED
        );
        
Why is  that?
The suggested method doesnt exist in the code

I would recommend we keep this method as it is useful and better than using the native DateInternal class and its methods.
E.g. (string) casting to the native format is super important IMO.

I see that https://github.com/cakephp/chronos/commit/3449ecabecd5f029bec3342376b6a4c788b227f8 dropped it as there were some issues with the native class and compatibility.

I have the same issue when trying to actually create any kind of improved wrapper around it.
Maybe instead of extending it, we could encapsulate it using the proxy pattern?

But according to e.g. https://github.com/pauci/datetime/blob/master/src/DateInterval.php it should work with extension.

dereuromark:
I sth like this viable?
https://github.com/cakephp/chronos/compare/2.x...2.x-interval

markstory:
> I have the same issue when trying to actually create any kind of improved wrapper around it.
Maybe instead of extending it, we could encapsulate it using the proxy pattern?

Given how thorny `datetime` can be, I think the proxy/decorator pattern is the only good option we have. We could include a `toNative()` or `toDateInterval` method so that working with other libraries is reasonable.

dereuromark:
Downside is that the types on methods wouldnt work, if something expects a native one.

markstory:
> Downside is that the types on methods wouldnt work, if something expects a native one.

Which is why we'd have a `toNative()` type method to provide compatibility with non-application code.

dereuromark:
Related to https://github.com/cakephp/chronos/pull/472 ?

dereuromark:
I opened a PR for discussion. Closing this then for now.

PR #495, opened 2026-01-28
Add ChronosInterval class

dereuromark:
## Summary

This PR adds a `ChronosInterval` class that wraps `DateInterval` using the decorator pattern, as suggested by @markstory in #444.

### Features

- **ISO 8601 Duration Formatting**: `__toString()` returns proper ISO 8601 format (e.g., `P1Y2M3DT4H5M6S`)
- **Factory Methods**:
  - `create(string $spec)` - Create from ISO 8601 spec
  - `createFromValues(...)` - Create from individual components
  - `instance(DateInterval $interval)` - Wrap an existing DateInterval
- **Native Interop**: `toNative()` returns the underlying `DateInterval` for compatibility
- **Convenience Methods**:
  - `totalSeconds()` - Approximate total seconds
  - `totalDays()` - Total days (exact if from diff())
  - `isNegative()` - Check if interval is inverted
  - `isZero()` - Check if interval has no duration
- **Property Access**: Proxies `y`, `m`, `d`, `h`, `i`, `s`, `f`, `invert`, `days`

### Example Usage

```php
// Create from spec
$interval = ChronosInterval::create('P1Y2M3D');
echo $interval; // "P1Y2M3D"

// Create from values
$interval = ChronosInterval::createFromValues(hours: 2, minutes: 30);
echo $interval; // "PT2H30M"

// Wrap existing DateInterval
$diff = $date1->diff($date2);
$interval = ChronosInterval::instance($diff);

// Use with native code
someFunctionExpectingDateInterval($interval->toNative());
```

## Discussion Points

This is a draft PR to gather feedback on the API design:

1. Should we add arithmetic methods (`add()`, `sub()`)? 
2. Should `ChronosInterval` be returned from `Chronos::diff()` instead of `DateInterval`?
3. Any other methods that would be useful?

Related to #444

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -6,6 +6,7 @@ Chronos provides a zero-dependency ``DateTimeImmutable`` extension, Date-only an
 * ``Cake\Chronos\Chronos`` extends ``DateTimeImmutable`` and provides many helpers.
 * ``Cake\Chronos\ChronosDate`` represents calendar dates unaffected by time or time zones.
 * ``Cake\Chronos\ChronosTime`` represents clock times independent of date or time zones.
+* ``Cake\Chronos\ChronosInterval`` wraps ``DateInterval`` with ISO 8601 formatting and convenience methods.
 * Only safe, immutable objects.
 * A pluggable translation system. Only English translations are included in the
   library. However, ``cakephp/i18n`` can be used for full language support.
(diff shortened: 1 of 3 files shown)


LordSimal:
1. Should we add arithmetic methods (add(), sub())?
Yes, I can definitely see it being usefull.

2. Should ChronosInterval be returned from Chronos::diff() instead of DateInterval?
For the next major, I'd say yes. Especially if we implement [`createFromDateString(string $datetime)` ](https://www.php.net/manual/en/dateinterval.createfromdatestring.php) as well and just map it to the `create()` method.

3. Any other methods that would be useful?
Other than `createFromDateString` I can see methods like `->next(Chronos::TUESDAY)` like we have in the Chronos class would be cool to adjust an interval to a desired range.

markstory (approved):
Looks good to me. The decorator based design avoids all the problems we've had with extending `datetime` classes in the past.

markstory:
> Should we add arithmetic methods (add(), sub())?

I think these are useful for combining intervals together.

> Should ChronosInterval be returned from Chronos::diff() instead of DateInterval?

Wouldn't this be a breaking change? I don't think we can make that switch outside of a major release. 

> Any other methods that would be useful?

Perhaps a method to convert the interval in to a `strottime()` compatible expression?

dereuromark:
@LordSimal Could you clarify what you meant by `->next(Chronos::TUESDAY)` style methods for intervals?

In `Chronos`, `next()` works from a specific date reference point. But intervals don't have an origin date, so I'm not sure how this would work semantically.

Some possible interpretations:
- Round interval UP to next full unit? (`P1Y2M15D` -> `P1Y3M`)
- Align/snap to calendar boundaries? (10 days -> 2 weeks)
- Something else?

Would appreciate an example of the use case you had in mind.

dereuromark:
Docs added.

dereuromark:
This would be the final piece then for the new minor release.

LordSimal:
I guess my usecase can just be achieved via generating the diff/interval with another chronos instance instead of modifiying the interval instance itself.

So forget about what I said. I haven't used intervals that often :D

LordSimal (approved):

Issue #406, opened 2023-08-15
Cannot construct a relative date that always points to the _next_ future date & time


Bilge:
If we construct a date at _midday, Tuesday_ with `new Chronos('Tue 12:0')`, but we are currently at 4pm tuesday, then this date will be 4 hours in the past. Let's suppose **we always want the constructed date to be in the future**. We could "fix" this by specifying `new Chronos('next Tue 12:0')`, which will point to 7 days ahead. However, if we are currently at 9am on a Tuesday, this will _still_ point 7 days ahead, instead of to midday on the current day, because the `next` specifier only considers the _date_ portion of the date/time compound and completely ignores the time. Therefore it is impossible to construct a relative date & time that always points to the next future occurrence of that specific date and time.

othercorey:
Chronos uses DateTimeImmutable for all parsing so we'd have to perform a post-parse adjustment to get this behavior.

dereuromark:
PR open. Please let us know if that would work.

PR #496, opened 2026-01-28
Add nextOccurrenceOf() and previousOccurrenceOf() methods

dereuromark:
## Summary

Adds two new methods to `Chronos` that solve the problem of constructing a relative date that always points to the next (or previous) future occurrence of a specific day *and* time.

### Problem

As described in #406, PHP's relative date parsing with `next` only considers the date, not the time:

```php
// If it's Tuesday 4pm:
new Chronos('Tue 12:00');       // Tuesday 12pm (in the past!)
new Chronos('next Tue 12:00');  // Next week Tuesday 12pm (always skips to next week)
```

There's no way to get "the next future occurrence of Tuesday 12pm" that returns today if it's before noon, or next week if it's after noon.

### Solution

Two new methods:

- `nextOccurrenceOf(int $dayOfWeek, int $hour, int $minute, int $second)` - Returns the next future occurrence, considering both day and time
- `previousOccurrenceOf(int $dayOfWeek, int $hour, int $minute, int $second)` - Returns the most recent past occurrence

```php
// If it's Tuesday 9am:
Chronos::now()->nextOccurrenceOf(Chronos::TUESDAY, 12, 0);
// → Today at 12pm (still in the future)

// If it's Tuesday 4pm:
Chronos::now()->nextOccurrenceOf(Chronos::TUESDAY, 12, 0);
// → Next Tuesday at 12pm (today's has passed)
```

### Behavior at exact time

When the current time matches exactly, `nextOccurrenceOf` returns next week and `previousOccurrenceOf` returns last week (strict comparison, not inclusive).

## Discussion Points

- Method naming: `nextOccurrenceOf` vs `nextOrSame` or something shorter?
- Should the time be at the exact second, or should it also consider `greaterThanOrEquals`?

Related to #406

## Test plan

- [x] Tests for same day before/after target time
- [x] Tests for different day
- [x] Tests for exact time edge case
- [x] Tests for seconds parameter
- [x] All existing tests pass (907 tests)
- [x] Code style passes

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -1619,6 +1619,89 @@ class Chronos extends DateTimeImmutable implements Stringable
     }
 
     /**
+     * Get the next occurrence of a given day of the week at a specific time.
+     *
+     * Unlike `next()`, this method considers both the day AND the time. If
+     * today is the target day and the specified time hasn't passed yet,
+     * it returns today at that time. Otherwise, it returns next week.
+     *
+     * This is useful when you need a relative date that always points to
+     * the next future occurrence of a specific day and time.
+     *
+     * ### Example
+     *
+     * ```
+     * // If it's Tuesday 9am, get "Tuesday 12pm" (today)
+     * // If it's Tuesday 4pm, get "Tuesday 12pm" (next week)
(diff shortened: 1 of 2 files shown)


Bilge:
Seems OK to me.

ADmad (approved):

PR #497, opened 2026-02-06
Cleanup for new major (required by Cake6)

dereuromark:
## Summary

Major release cleanup for Chronos 4.0 / CakePHP 6.0 compatibility:

- Remove the `$range` parameter from `toQuarter()` which was deprecated in 3.3.0
  - Users should use `toQuarterRange()` for quarter range functionality
- Widen `$toStringFormat` property type from `string` to `array|string|int`
  - This allows CakePHP I18n classes to rename `$_toStringFormat` to `$toStringFormat` without type conflicts
  - Enables CakePHP upgrade tool to apply PSR naming conventions consistently
- Remove outdated PHPStan ignore for `DateTimeImmutable::createFromTimestamp()` (now available in PHP 8.4)

Refs: https://github.com/cakephp/upgrade/pull/386

Note:
The type widening is backwards compatible because CakePHP 5 I18n classes use their own $_toStringFormat property and override all related methods - they don't touch Chronos's property at all. The widening just enables       
  CakePHP 6 to do the PSR naming cleanup.

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -163,9 +163,12 @@ class Chronos extends DateTimeImmutable implements Stringable
     /**
      * Format to use for __toString method when type juggling occurs.
      *
-     * @var string
+     * The widened type allows subclasses (like CakePHP I18n classes) to use
+     * IntlDateFormatter constants while maintaining backward compatibility.
+     *
+     * @var array|string|int
      */
-    protected static string $toStringFormat = self::DEFAULT_TO_STRING_FORMAT;
+    protected static array|string|int $toStringFormat = self::DEFAULT_TO_STRING_FORMAT;
 
     /**
      * Days of weekend
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Run phpcs
         if: always()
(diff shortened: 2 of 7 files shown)


src/Chronos.php
```diff
@@ -163,9 +163,12 @@ class Chronos extends DateTimeImmutable implements Stringable
...
      * Format to use for __toString method when type juggling occurs.
      *
-     * @var string
+     * The widened type allows subclasses (like CakePHP I18n classes) to use
+     * IntlDateFormatter constants while maintaining backward compatibility.
+     *
+     * @var array|string|int
      */
-    protected static string $toStringFormat = self::DEFAULT_TO_STRING_FORMAT;
+    protected static array|string|int $toStringFormat = self::DEFAULT_TO_STRING_FORMAT;
```
LordSimal, line 171: Isn't this considered a breaking change for non cakephp apps using this?

dereuromark: Yes, thats BC breaking why this is targeting 4x, see milestone and PR descr.
Same as deprecation removal etc is.

But no, not for Cake Apps, explained above.

LordSimal: Then we should rename 3.next to 4.x, thats why i was confused^^

PR #504, opened 2026-03-14
4.x: Auto-append pipe modifier in createFromFormat

dereuromark:
Attempt of https://github.com/cakephp/chronos/pull/494 redo of "deterministic behavior"

## Summary

Auto-append `|` modifier in `createFromFormat()` for deterministic behavior.

### The Change

Unlike PHP's native `DateTimeImmutable::createFromFormat()`, this method now automatically appends the `|` modifier if no reset modifier (`|` or `!`) is present.

This ensures that unparsed components are reset to zero instead of being filled from the current time.

The 4.x philosophy: Always deterministic, everywhere                                                                                                                     

### Examples

```php
// Before (PHP native behavior): time filled from current system time
// After (Chronos 4.x): time is 00:00:00.000000
Chronos::createFromFormat('Y-m-d', '2024-03-14');

// Before: seconds filled from current system time
// After: seconds is 00
Chronos::createFromFormat('Y-m-d H:i', '2024-03-14 12:30');

// Missing date components use Unix epoch (1970-01-01)
Chronos::createFromFormat('H:i:s', '12:30:45');
// Result: 1970-01-01 12:30:45
```

This is what most users intuitively expect. When you parse just a date, you get midnight - not "midnight plus whatever seconds happen to be on the clock."                                                                                                                                                 
                                                                                                                                                          
The only "BC break" scenario is someone who intentionally wanted:                                                                                       
```php
  // "Give me this date but with current time" (unusual use case)                                                                                         
  $d = Chronos::createFromFormat('Y-m-d', '2024-03-14');                                                                                                  
  // They expected 2024-03-14 14:32:45 (current time)                                                                                                     
  // Now they get 2024-03-14 00:00:00                                                                                                                     
```                                                                                                                                                          
That's a rare/unusual pattern. If someone actually needs current time for missing components, they'd explicitly do:                                     
```php
  $d = Chronos::parse('2024-03-14'); // This keeps current time                                                                                           
  // or construct it explicitly    
```

### Benefits

- **Deterministic**: No more flaky tests due to time-dependent behavior
- **Intuitive**: Missing = zero is what most users expect
- **No workarounds needed**: No need to add `|` to format strings manually

### Migration

Test Code that relied on missing components being filled from current time will need adjustment.
But this "test" BC break is more theoretical than practical:                                                                                                     
                                                                                                                                                          
  1. People who wanted zeros → already using setTestNow() or adding | manually                                                                            
  2. People who wanted current time → this is the edge case that breaks, but it's likely unintentional/accidental rather than deliberate

### DIFF
--- a/src/Chronos.php
+++ b/src/Chronos.php
@@ -668,6 +668,12 @@ class Chronos extends DateTimeImmutable implements Stringable
     /**
      * Create an instance from a specific format
      *
+     * Unlike PHP's native DateTimeImmutable::createFromFormat(), this method
+     * automatically appends the `|` modifier if no reset modifier (`|` or `!`)
+     * is present. This ensures that unparsed components are reset to zero
+     * instead of being filled from the current time, providing more predictable
+     * and deterministic behavior.
+     *
      * @param string $format The date() compatible format string.
      * @param string $time The formatted date string to interpret.
      * @param \DateTimeZone|string|null $timezone The DateTimeZone object or timezone name the new instance should use.
@@ -679,6 +685,12 @@ class Chronos extends DateTimeImmutable implements Stringable
         string $time,
         DateTimeZone|string|null $timezone = null,
     ): static {
+        // Auto-append | modifier if no reset modifier is present
+        // This ensures unparsed components are zero instead of current time
+        if (!str_contains($format, '|') && !str_contains($format, '!')) {
+            $format .= '|';
+        }
+
         if ($timezone !== null) {
             $dateTime = parent::createFromFormat($format, $time, $timezone ? static::safeCreateDateTimeZone($timezone) : null);
         } else {
--- a/tests/TestCase/DateTime/CreateFromFormatTest.php
+++ b/tests/TestCase/DateTime/CreateFromFormatTest.php
@@ -29,6 +29,49 @@ class CreateFromFormatTest extends TestCase
         $this->assertTrue($d instanceof Chronos);
     }
 
+    public function testCreateFromFormatMissingTimeIsZero()
+    {
+        // Missing time components should be zero, not current time
+        $d = Chronos::createFromFormat('Y-m-d', '2024-03-14');
+        $this->assertDateTime($d, 2024, 3, 14, 0, 0, 0);
+        $this->assertSame(0, $d->micro);
+    }
+
+    public function testCreateFromFormatMissingSecondsIsZero()
+    {
+        // Missing seconds should be zero
+        $d = Chronos::createFromFormat('Y-m-d H:i', '2024-03-14 12:30');
+        $this->assertDateTime($d, 2024, 3, 14, 12, 30, 0);
+    }
+
+    public function testCreateFromFormatMissingDateIsEpoch()
+    {
+        // Missing date components should be Unix epoch (1970-01-01)
+        $d = Chronos::createFromFormat('H:i:s', '12:30:45');
+        $this->assertDateTime($d, 1970, 1, 1, 12, 30, 45);
+    }
+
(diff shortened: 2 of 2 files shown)


dereuromark:
## Approach Comparison

### Behavior Matrix

| Scenario | PHP Native | #494 (testNow) | #504 (auto-pipe) | Carbon |
|----------|------------|----------------|------------------|--------|
| No testNow, `Y-m-d` | current time | current time | 00:00:00 | current time |
| testNow=14:30:45, `Y-m-d` | current time | 14:30:45 | 00:00:00 | 14:30:45 |
| No testNow, `Y-m-d H:i` | current seconds | current seconds | :00 | current seconds |
| testNow, `Y-m-d H:i` | current seconds | testNow seconds | :00 | testNow seconds |
| Format with `\|` | zero | zero | zero | zero |
| Format with `!` | zero | zero | zero | zero |

### Feature Comparison

| Aspect | #494 (testNow) | #504 (auto-pipe) |
|--------|----------------|------------------|
| **Test determinism** | Yes (with setTestNow) | Yes (always) |
| **Production determinism** | No | Yes |
| **Matches Carbon** | Yes | No |
| **Matches PHP native** | Yes (without testNow) | No |
| **Test = Production behavior** | No | Yes |
| **Format parsing required** | Yes (complex) | No (simple) |
| **Edge cases to handle** | `U`, `U.u`, etc. | None |

### Pros and Cons

#### #494 (testNow approach)

**Pros:**
- Matches Carbon behavior - familiar to users migrating
- Less BC break in production (native PHP behavior preserved when no testNow)
- Only changes behavior in test context

**Cons:**
- Production still has footgun (missing = random current time)
- Tests behave differently than production (could mask bugs)
- Complex implementation (must parse format specifiers)
- Edge cases like `U` need special handling (#502)
- More code to maintain

#### #504 (auto-append pipe)

**Pros:**
- Deterministic everywhere - no surprises in tests OR production
- Test behavior = production behavior (what you test is what you get)
- Simple implementation (just append `|` if no reset modifier)
- No edge cases - PHP's `|` modifier handles all formats correctly
- Less code to maintain

**Cons:**
- BC break for anyone relying on "missing = current time" (rare)
- Different from Carbon behavior
- Different from PHP native behavior

## Implementation Complexity

### #494 (testNow) - Format Parsing Required

```php
// Must detect which components are present in format
// Handle escaped characters (\Y, \\, etc.)
// Special cases: U, U.u, timezone formats
// ~50+ lines of parsing logic
```

### #504 (auto-pipe) - Simple String Check

```php
if (!str_contains($format, '|') && !str_contains($format, '!')) {
    $format .= '|';
}
// 3 lines, delegates complexity to PHP
```

## Real-World Impact

### Who is affected by #504 (auto-pipe)?

Code that **intentionally** does this:
```php
// "Give me this date with current system time"
$d = Chronos::createFromFormat('Y-m-d', $dateString);
// Expected: 2024-03-14 14:32:45 (current time)
// With #504: 2024-03-14 00:00:00
```

This is a rare/unusual pattern. More common is users being surprised that they got random seconds.

### Who is affected by #494 (testNow)?

Code that uses `createFromFormat` without `setTestNow`:
- No change (native PHP behavior)

Code that uses `createFromFormat` WITH `setTestNow`:
- Missing components now come from testNow instead of current time
- This is usually the desired behavior for tests

src/Chronos.php
```diff
@@ -671,6 +668,12 @@ public static function createFromTime(
     /**
      * Create an instance from a specific format
      *
+     * Unlike PHP's native DateTimeImmutable::createFromFormat(), this method
```
ADmad, line 671: This sounds like a sensible default 👍

ADmad (approved):

PR #509, opened 2026-03-15
fix CS

LordSimal:


dereuromark:
It looks like the docs cannot deploy for a while on 3.x, is that possible?

PR #512, opened 2026-03-22
Improve documentation coverage

dereuromark:
## Summary

Addresses documentation gaps identified in the codebase:

- Expand README with ChronosTime, `setTestNow()`, and ClockFactory (PSR-20) examples
- Improve ClockFactory class and method documentation
- Add detailed docs for `createInterval()` explaining time unit rollover behavior
- Improve `average()` docs with example showing midpoint calculation
- Improve `diffFiltered()` docs with callback signature and usage example
- Improve `nthOfMonth/Quarter/Year` docs explaining the false return case

The README now covers more of the public API including commonly used testing utilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,64 @@ echo $today->modify('+3 hours');
 
 Like instances of `Chronos`, `ChronosDate` objects are also *immutable*.
 
+# Time-only Values
+
+When you need to work with just times (without dates), use `ChronosTime`:
+
+```php
+use Cake\Chronos\ChronosTime;
+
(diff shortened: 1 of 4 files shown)


LordSimal:
The only thing I can mention here is the fact, that the new Vitepress docs in `docs` should also probably reflect this change (if it not already does)

dereuromark:
Yes, I plan to make a full follow up here after we finished it all with the minor release and stuff.

dereuromark:
Feel free to merge if all good.

PR #515, opened 2026-03-26
Return ChronosInterval from Chronos::diff()

### LINKED ISSUE #511: Return ChronosInterval from Chronos::diff()
Following discussion in #495, this is a tracking issue for breaking changes to consider for 4.x.

### Return `ChronosInterval` from `Chronos::diff()`

Currently `Chronos::diff()` returns a native `DateInterval`. In 4.x, it should return `ChronosInterval` instead to provide:

- ISO 8601 duration formatting via `__toString()`
- Convenience methods like `totalSeconds()`, `totalDays()`, `isZero()`, `isNegative()`
- Arithmetic methods `add()`, `sub()`
- `toDateString()` for strtotime-compatible output
- `toNative()` for backwards compatibility when a `DateInterval` is needed

### Migration Path

Users who need a `DateInterval` can call `->toNative()` on the result:

```php
// Before (3.x)
$interval = $date1->diff($date2);

// After (4.x) - if DateInterval is needed
$interval = $date1->diff($date2)->toNative();

// After (4.x) - using new features
$interval = $date1->diff($date2);
echo $interval; // "P1Y2M3D"
echo $interval->totalDays();
```

## References

- #495 - ChronosInterval implementation
- #444 - Original discussion

dereuromark:
## Summary

- Changes `Chronos::diff()` and `ChronosDate::diff()` to return `ChronosInterval` instead of `DateInterval`
- Adds the `ChronosInterval` class (merged from 3.x branch)
- Updates `fromNow()` to return `ChronosInterval`

This provides:

- ISO 8601 duration formatting via `__toString()`
- Convenience methods like `totalSeconds()`, `totalDays()`, `isZero()`, `isNegative()`
- Arithmetic methods `add()`, `sub()`
- `toDateString()` for strtotime-compatible output
- `toNative()` for backwards compatibility when a `DateInterval` is needed

## Migration Path

Users who need a `DateInterval` can call `->toNative()` on the result:

```php
// Before (3.x)
$interval = $date1->diff($date2);

// After (4.x) - if DateInterval is needed
$interval = $date1->diff($date2)->toNative();

// After (4.x) - using new features
$interval = $date1->diff($date2);
echo $interval; // "P1Y2M3D"
echo $interval->totalDays();
```

## Note

Good news: No changes required in CakePHP core! 

PHPStan reports an error about covariant return types - this is expected for this BC break. The baseline will need to be updated.

Closes #511

### DIFF
--- a/docs/.vitepress/toc_en.json
+++ b/docs/.vitepress/toc_en.json
@@ -5,6 +5,7 @@
       "collapsed": false,
       "items": [
         { "text": "Introduction", "link": "/index" },
+        { "text": "4.x Migration Guide", "link": "/4-x-migration-guide" },
         { "text": "3.x Migration Guide", "link": "/3-x-migration-guide" },
         { "text": "API", "link": "https://api.cakephp.org/chronos" }
       ]
--- a/docs/en/4-x-migration-guide.md
+++ b/docs/en/4-x-migration-guide.md
@@ -1,0 +1,28 @@
+# 4.x Migration Guide
+
+Chronos 4.x contains breaking changes that could impact your application. This
+guide provides an overview of the breaking changes made in 4.x
+
+## diff() and fromNow() return ChronosInterval
+
+`Chronos::diff()`, `ChronosDate::diff()` and `Chronos::fromNow()` now return a
+`ChronosInterval` instead of a `DateInterval`. `Chronos::fromNow()` previously
+returned `DateInterval|false`; it now always returns a `ChronosInterval`.
+
+`ChronosInterval` decorates the native `DateInterval` and exposes the same
+properties (`y`, `m`, `d`, `h`, `i`, `s`, `f`, `invert`, `days`), so most code
+keeps working unchanged. However, it does **not** extend `DateInterval`: code
+that type-hints `DateInterval` or relies on `instanceof DateInterval` against
+the result must call `->toNative()` to get the underlying `DateInterval`:
+
+```php
+// Before (3.x)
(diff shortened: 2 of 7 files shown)


dereuromark:
For downstream users:                                                                                              

  Anyone using $chronos->diff() with type hints will need updates:                                                   
```php
// If they have:                                                                                                   
function process(DateInterval $interval) { ... }                                                                   
                                                                                                                     
// They need:                                                                                                      
function process(DateInterval|ChronosInterval $interval) { ... }                                                   
// Or:                                                                                                             
$date->diff($other)->toNative()  // to get DateInterval                                                            
```


## CI Status

All 954 tests pass. The CI failures are due to:

1. **PHPStan** - Expected error about covariant return type:
   > Return type ChronosInterval of method Chronos::diff() is not covariant with tentative return type DateInterval of method DateTimeInterface::diff()

2. **PHPUnit (PHP 8.3+)** - The repo has `failOnDeprecation="true"` in phpunit.xml.dist, so the PHP deprecation notice about incompatible return types causes tests to fail even though they all pass.

Both issues are expected for this BC break. We will need to decide how to handle:
- Update phpstan.neon to ignore this specific error
- Either update phpunit.xml.dist or add baseline for the deprecation

LordSimal:
Docs are missing, other than that, it looks good.

dereuromark:
Done.

jamisonbryant (approved):
Looks good to me. Docs look good too. Should we add a “since” admonition like the book does or is the migration guide sufficient?

PR #523, opened 2026-04-14
Add arithmetic, diff and modify methods to ChronosTime

### LINKED ISSUE #522: Missing ChronosTime differences and modifiers methods
With version 3 [ChronosTime](https://github.com/cakephp/chronos/blob/3.x/src/ChronosTime.php) lost all differences and modifiers methods like:
```php
public function modify(string $modifier): static;
public function diff(ChonosTime $target, bool $absolute = false): DateInterval;
public function addHours(int $value): static;
public function subHours(int $value): static;
public function addMinutes(int $value): static;
public function subMinutes(int $value): static;
public function addSeconds(int $value): static;
public function subSeconds(int $value): static;

public function closest(ChonosTime $first, ChonosTime $second, ChonosTime ...$others): static;
public function farthest(ChonosTime $first, ChonosTime $second, ChonosTime ...$others): static;
public function min(?ChonosTime $other = null): static;
public function max(?ChonosTime $other = null): static;
public function average(?ChonosTime $other = null): static;
public function wasWithinLast(string|int $timeInterval): bool;
public function isWithinNext(string|int $timeInterval): bool;

public function diffFiltered(DateInterval $interval, callable $callback, ?ChonosTime $other = null, bool $absolute = true, int $options = 0): int;
public function diffInHoursFiltered(callable $callback, ?ChonosTime $other = null, bool $absolute = true, int $options = 0): int;
public function diffInHours(?ChonosTime $other = null, bool $absolute = true): int;
public function diffInMinutes(?ChonosTime $other = null, bool $absolute = true): int;
public function diffInSeconds(?ChonosTime $other = null, bool $absolute = true): int;
public function secondsSinceMidnight(): int;
public function diffForHumans(?ChonosTime $other = null, bool $absolute = false): string;
```

Furthermore, in [README.md](https://github.com/cakephp/chronos/tree/3.x?tab=readme-ov-file#time-only-values), the "Time-only Values" section is inconsistent, once it describes unexisting methods, like:
```php
$later = $time->addHours(2)->addMinutes(15);
```

dereuromark:
## Summary

Fixes #522.

`ChronosTime` was introduced in 3.x as a brand-new standalone class and — unlike `Chronos` and `ChronosDate`, which compose behavior via traits (`ModifierTrait`, `DifferenceTrait`, …) — it shipped without any arithmetic, diff or modifier helpers. That left a time-only value effectively read-only, and broke the README's "Time-only Values" example:

~~~php
$later = $time->addHours(2)->addMinutes(15);
~~~

This PR adds the missing API directly to `ChronosTime`. Everything operates on the existing internal `$ticks` counter (microseconds since midnight) and wraps via the existing `mod()` helper, so no `DateTimeImmutable` round-trips are needed for arithmetic.

## Changes

- ``addHours``/``subHours``, ``addMinutes``/``subMinutes``, ``addSeconds``/``subSeconds``
- ``modify(string \$modifier)`` — restricted by regex to time-unit modifiers (``hour``/``minute``/``second``/``microsecond``, signed integer, optionally chained). Date-unit modifiers (``+1 day``, ``next monday``, etc.) throw ``InvalidArgumentException`` rather than silently returning the wrong result.
- ``diff(ChronosTime \$target, bool \$absolute = false)`` returning a ``DateInterval`` — default matches ``DateTimeInterface::diff``.
- ``diffInHours``/``diffInMinutes``/``diffInSeconds`` — ``\$absolute = true`` by default (Chronos convention).
- ``secondsSinceMidnight``
- ``closest``/``farthest``/``min``/``max``

## Wraparound semantics

All arithmetic is modulo one day, matching the existing setter behavior (``setHours(24) === setHours(0)``):

- ``23:00 + 2h → 01:00``
- ``00:00 − 2h → 22:00``

Each method's docblock calls this out explicitly.

## Not included (and why)

- ``diffForHumans``, ``wasWithinLast``, ``isWithinNext`` — these imply an instant on a timeline, not a wall-clock time. Semantics are ambiguous for a date-less value, so I left them out rather than fake them. Happy to add if you want them.
- ``average`` — ambiguous under wraparound (naive mean of ``23:00`` and ``01:00`` gives ``12:00``, not midnight). Left out pending a decision on desired semantics.
- ``diffFiltered``/``diffInHoursFiltered`` — these depend on ``CarbonPeriod``-style iteration over dates; doesn't map cleanly to a time-only value.

## Notes

- Unrelated preexisting phpstan issue on ``Chronos.php`` (``DateTimeImmutable::createFromTimestamp`` ignore pattern no longer matching under PHP 8.4) is present on ``3.next`` already and not touched here.

### DIFF
--- a/src/ChronosTime.php
+++ b/src/ChronosTime.php
@@ -445,24 +445,26 @@ class ChronosTime implements Stringable
      */
     public function modify(string $modifier): static
     {
-        $pattern = '/^(?:\s*[+-]?\s*\d+\s*(?:hour|minute|second|microsecond)s?\s*)+$/i';
-        if (!preg_match($pattern, $modifier)) {
+        if (!preg_match('/^(?:\s*[+-]?\s*\d+\s*(?:hour|minute|second|microsecond)s?\s*)+$/i', $modifier)) {
             throw new InvalidArgumentException(
                 sprintf('Modifier `%s` is not a valid ChronosTime modifier.', $modifier),
             );
         }
 
-        $base = new DateTimeImmutable('1970-01-01 00:00:00', new DateTimeZone('UTC'));
-        $modified = $base->modify($modifier);
-        if ($modified === false) {
-            throw new InvalidArgumentException(
-                sprintf('Unable to apply modifier `%s` to ChronosTime.', $modifier),
-            );
-        }
+        $unitTicks = [
+            'hour' => self::TICKS_PER_HOUR,
+            'minute' => self::TICKS_PER_MINUTE,
+            'second' => self::TICKS_PER_SECOND,
+            'microsecond' => self::TICKS_PER_MICROSECOND,
+        ];
 
-        $deltaSeconds = $modified->getTimestamp() - $base->getTimestamp();
-        $deltaMicros = (int)$modified->format('u') - (int)$base->format('u');
-        $deltaTicks = $deltaSeconds * self::TICKS_PER_SECOND + $deltaMicros * self::TICKS_PER_MICROSECOND;
+        preg_match_all('/([+-]?\s*\d+)\s*(hour|minute|second|microsecond)s?/i', $modifier, $matches, PREG_SET_ORDER);
+
+        $deltaTicks = 0;
+        foreach ($matches as $match) {
+            $value = (int)preg_replace('/\s+/', '', $match[1]);
+            $deltaTicks += $value * $unitTicks[strtolower($match[2])];
+        }
 
         return $this->addTicks($deltaTicks);
     }
--- a/src/ChronosTime.php
+++ b/src/ChronosTime.php
@@ -20,7 +20,6 @@ use DateTimeInterface;
 use DateTimeZone;
 use InvalidArgumentException;
 use Stringable;
-use Throwable;
 
 /**
  * @phpstan-consistent-constructor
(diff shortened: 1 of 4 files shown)


src/ChronosTime.php
```diff
@@ -351,6 +352,305 @@ public function endOfHour(): static
...
+        $modified = $base->modify($modifier);
+        if ($modified === false) {
+            throw new InvalidArgumentException(
+                sprintf('Unable to apply modifier `%s` to ChronosTime.', $modifier),
+            );
+        }
+
+        $deltaSeconds = $modified->getTimestamp() - $base->getTimestamp();
+        $deltaMicros = (int)$modified->format('u') - (int)$base->format('u');
+        $deltaTicks = $deltaSeconds * self::TICKS_PER_SECOND + $deltaMicros * self::TICKS_PER_MICROSECOND;
```
LordSimal: can this get to a point where we overflow? I know basically all our systems now are 64 bit, so the chance of that happeneing is very slim, but there are surely some systems out there which still use 32 bit PHP and therefore this can lead to a Y2038 problem, right?

dereuromark: Good point! We don't need DateTimeImmutable at all in this method. 

On 32-bit PHP, `(new DateTimeImmutable('1970-01-01'))->modify('+X')->getTimestamp()` would overflow once the delta exceeded 2^31 seconds (~68 years, anything >~596k hours / >~2.1B seconds).

Fixed in 83bec6d by dropping the `DateTimeImmutable` round-trip entirely — the regex already constrains input to `[sign] int unit`, so I just walk the matches and accumulate ticks directly using the existing `TICKS_PER_*` constants. No epoch involvement, no `getTimestamp()`, immune to the platform `int` width. Added a regression test (`testModifyDoesNotOverflowOn32Bit`) using `+2400000 hours` (~8.6B seconds, well past 2^31) plus a compound-modifier test for good measure.

LordSimal (approved):

PR #533, opened 2026-09-16
Fix ChronosDate ordinal relative modifiers

### LINKED ISSUE #532: ChronosDate::modify() fails on 'second' relative keywords
### Description
Calling `modify()` on a `ChronosDate` instance with relative expressions such as `'second monday of this month'` throws an error.

### Steps to Reproduce
```php
use Cake\Chronos\ChronosDate;

$date = (new ChronosDate('2026-09-01'))->modify('second monday of this month');
```

### Expected Behavior
The `ChronosDate` instance is modified to the second Monday of the month.

### Actual Behavior
Throws an exception:
> Cannot modify date objects by time values

### Environment
- cakephp/chronos: 3.5.1
- PHP: 8.5

dereuromark:
## Summary

- parse relative modifiers before rejecting time components, so ordinal expressions such as `second monday of this month` remain valid
- reject uppercase, abbreviated, fractional, and wrapped-day time modifiers
- reject negative time components passed to `add()` and `sub()`
- preserve timestamp modifiers and date arithmetic across time zones that skip midnight

## Tests

- `composer test`
- `composer cs-check`
- `composer stan` (fails on the existing unmatched `DateTimeImmutable::createFromTimestamp()` ignore under PHP 8.5)

Closes #532


src/ChronosDate.php
```diff
@@ -338,7 +338,21 @@ public function sub(DateInterval $interval): static
      */
     public function modify(string $modifier): static
     {
-        if (preg_match('/hour|minute|second/', $modifier)) {
+        $parsed = date_parse($modifier);
+        $relative = $parsed['relative'] ?? [];
+        $timestampSeconds = 0;
+        if (preg_match('/(?:^|\s)@(-?\d+)(?:\.\d+)?(?:\s|$)/', $modifier, $matches)) {
```
markstory, line 344: Would it be simpler to check `native` after the `modify()` call and see if the new value has seconds/milliseconds/microseconds set?

